PR #369, opened 2026-05-31
[codex] Fix stale WG coordinator identity refs

### LINKED ISSUE #299: Fix WG coordinator peer discovery when legacy identity paths are stale
## Problem

From a verified WG coordinator replica, `list-peers-lean` should include coordinators from other workgroups and the synthetic Root Agent peer (`agentscommander://root-agent`). In the current `AgentsCommander_ac` workgroup replica, discovery only returns peers from the same WG and emits:

```text
Warning: no coordinator found for WG 'wg-2-dev-team', showing all replicas
```

## Root Cause

`resolve_wg_coordinator_replica` currently relies on `std::fs::canonicalize()` for both the team `coordinator` ref and the replica `identity` ref. `canonicalize()` requires the path to exist. Legacy configs may contain absolute or deep relative paths tied to a previous workspace folder name, so those paths fail to resolve after a workspace rename or relocation.

When coordinator identity verification fails, the caller is not recognized as coordinator, so `discover_wg_peers` skips:

- other WG coordinator peers
- the synthetic Root Agent peer

## Requirements

- Preserve identity-based authorization: never trust the replica directory name (`__agent_*`) as coordinator proof.
- Add a tolerant comparison path for legacy/stale identity refs so coordinator verification can succeed when refs point to the same Matrix Agent ID or equivalent normalized identity.
- Keep spoofing protections intact: a replica must still declare an identity matching the team coordinator identity.
- Ensure `list-peers` and `list-peers-lean` return the same peer set.

## Acceptance Criteria

- Running `list-peers-lean` from a verified WG coordinator includes other WG coordinators when present.
- Running `list-peers-lean` from a verified WG coordinator includes `agentscommander://root-agent`.
- Non-coordinator replicas do not gain Root Agent or cross-WG coordinator visibility.
- Existing spoofing regression tests continue to pass.
- Add Rust and/or CLI regression coverage for stale legacy identity paths.

mblua:
## Summary

Fixes #299.

This updates WG coordinator resolution so legacy or stale identity refs can still match when they point to the same Matrix Agent identity, even if `std::fs::canonicalize()` cannot resolve the old path on disk.

## Root Cause

`resolve_wg_coordinator_replica` previously required both the team `coordinator` ref and each replica `identity` ref to canonicalize successfully. After a workspace rename or relocation, legacy refs can point at a path that no longer exists, so coordinator verification failed and `list-peers`/`list-peers-lean` omitted cross-WG coordinators and the synthetic Root Agent peer.

## Changes

- Add a stable identity comparison key that canonicalizes existing paths and falls back to no-disk logical path normalization for stale refs.
- Keep authorization identity-based by comparing the team coordinator ref against the replica-declared `identity`, never the `__agent_*` directory name.
- Add regression coverage for stale absolute refs, stale relative refs, logical path normalization, case-insensitive comparison, and spoofed stale identities.

## Validation

- `rustfmt --edition 2021 --check src\config\teams.rs`
- `cargo test --lib resolve_wg_coordinator_replica`
- `cargo test --lib identity_compare_key`
- `cargo test --lib logical_path_resolve`
- `cargo test --lib verified_wg_coordinator_target`
- `cargo test --lib build_root_agent_synthetic_peer`
- `cargo test --lib discover_root_coordinator_peers_from`
- `cargo test --lib` from `src-tauri` after rebasing onto current `main` (`822 passed, 12 ignored`)

Also verified live CLI behavior in the workgroup: a verified coordinator sees other WG coordinators plus `agentscommander://root-agent`, while a non-coordinator sees only same-WG peers. `list-peers` and `list-peers-lean` returned matching peer sets.

### DIFF
--- a/src-tauri/src/config/teams.rs
+++ b/src-tauri/src/config/teams.rs
@@ -213,10 +213,104 @@ pub struct WgCoordinatorReplica {
     pub replica_dir: PathBuf,
 }
 
-fn canonical_path_string(path: &Path) -> Option<String> {
-    let canon = std::fs::canonicalize(path).ok()?;
-    let s = canon.to_string_lossy().to_string();
-    Some(s.strip_prefix(r"\\?\").unwrap_or(&s).to_string())
+/// Normalize a path string to a stable comparison form.
+///
+/// - Strip the Windows extended-length (`\\?\`) prefix.
+/// - Replace `\` with `/`.
+/// - Lowercase (case-insensitive comparison; matches the convention used by
+///   `cli/list_peers::norm_path`).
+/// - Trim trailing `/`.
+fn normalize_path_for_compare(s: &str) -> String {
+    let stripped = s.strip_prefix(r"\\?\").unwrap_or(s);
+    stripped
+        .replace('\\', "/")
+        .to_lowercase()
+        .trim_end_matches('/')
+        .to_string()
+}
+
+/// Logical resolution of `.` and `..` components in a path string (no disk
+/// access). Used as a fallback when `std::fs::canonicalize()` fails because
+/// the target doesn't exist — e.g., legacy identity refs that point to a
+/// previous workspace folder name (see #299).
+///
+/// Returns a forward-slash string. `..` components above the root of an
+/// absolute path are dropped; `..` components at the front of a purely
+/// relative path are preserved.
+fn logical_path_resolve(path: &Path) -> String {
+    use std::path::Component;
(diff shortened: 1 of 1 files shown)


mblua:
Superseded by #370. This PR used a branch name that failed the repository branch-name validation workflow (`codex/...` instead of `<type>/<issue>-<slug>`). The replacement PR uses `fix/299-wg-coordinator-stale-identity`.

Issue #1341, opened 2026-08-14
Auto-update prompt never fires in a fresh install with a registered agent


mblua:
# Auto-update prompt never fires in a real fresh install (registered agent, empty policy map)

## Situation
On a fresh standalone install (main 9196ea97, #1335 included), the SI/NO auto-update prompt never appears, in a case where it should:

- Exe: `agentscommander_standalone.exe` (build of main 9196ea97 — verified: `agent_update` strings present, `claude --update`/`pi update`/`codex update` present in the binary).
- Config dir `D:\0_repos\test\.agentscommander_standalone\`: `settings.json` has `agents: [{label: "Claude Code", command: "claude"}]`, `agentAutoUpdateByCommand: {}` (absent → should prompt).
- No `coding-agents/` dir in the config dir; app log shows "agents.json absent; serving embedded default catalog" on every boot (6 boots).
- `projectPaths: []` (no project registered).
- App log has ZERO `[agent-update]` lines across all boots (even the debug early-return is absent, though default log level may filter it).

Expected: first boot with claude registered + empty policy map → prompt for command "claude" (registered filter #1335 passes it: claude IS in settings.agents). Observed: no prompt, no update, nothing.

## Why (hypotheses to check, evidence-driven)
1. `run_startup_updates` task not spawned or aborted early in the fresh-install path (lib.rs setup ordering, single-instance, restore, or a conditional that skips it).
2. The gate/prompt runs but the emitted event/snapshot never reaches the sidebar overlay (F3 snapshot path) in a no-project install.
3. `load_catalog_for_settings` in this shape returns a catalog WITHOUT update_commands (embedded default served — but is the served copy the one with updateCommands, or an older cached shape?).
4. The prompt fired but auto-dismissed (60s timeout default No) and the user missed it / it raced with first-run onboarding.

## How
Diagnose with the real exe (logLevel debug, watch `[agent-update]` lines; if needed add temporary diagnostic logging — decide from evidence, not hunches). Fix whatever the evidence identifies. Unit/integration coverage for the found cause.

## Out of scope
The per-agent Auto-update dropdown (separate issue).

mblua:
[...]

Issue #1327, opened 2026-08-13
Coding-agent auto-update flow: run updateCommands at startup (splash + per-command policy)


mblua:
# Coding-agent auto-update flow: run updateCommands at startup (splash + per-command policy)

## Situation
The catalog now carries `updateCommands` per coding agent (claude `["claude --update"]`, pi `["pi update"]`, codex `["codex update"]`; others empty), `autoUpdate` (stable default, false), and per-user policy maps in `AppSettings` (`agentAutoUpdate`, `agentUpdateDontAskAgain`). All of it is inert — nothing reads or executes it. The update-check UI and flow were explicitly out of scope of #1318.

## Why
The user wants the update flow implemented with the simplest sane semantics: **no version detection/comparison at all** — the update command itself already decides whether there is something new (`pi update`, `claude --update` self-check). The app only needs to run the commands at the right time, with the right policy and UX.

## Decided behavior (user-confirmed 2026-08-13)
1. **Policy granularity: per `command` string, not per agent key.** Multiple coding-agent profiles can use the same real software (e.g. two profiles both `command: "claude"`); the software is the unit of update. The first-time question and the stored setting are keyed by the command string.
2. **First-time question (per command):** when a new command is first detected among registered coding agents, ask once: "¿Querés que al arranque se intente actualizar automáticamente el coding agent X?" SI / NO. Answer is stored and **never asked again for that command**.
3. **Default: No** (matches `autoUpdate: false` in the catalog).
4. **At startup:** for every registered agent whose command is enabled for auto-update, execute its `updateCommands` sequence **in parallel across commands** (each agent's own sequence stays ordered). This happens **before any agent session can open** (avoids conflicts: updating the CLI that a session would launch).
5. **UX while updating:** the app shows a **splash / indicator** ("actualizando agents...") and waits; sessions are not usable until updates finish.
6. **Errors:** if an update command fails, show the app's standard **red error message** (the established error UI), per failing command; do not block the rest. Continue startup afterwards.
7. Commands with empty `updateCommands` are skipped. A command with no update command never prompts.

## How (scope for the plan)
1. Startup orchestration: where in the boot sequence the update run sits (before sessions become available), parallel execution (e.g. `futures`/`tokio::join!` or a task set), the wait/splash gate.
[...]

Issue #1343, opened 2026-08-14
Restore deterministic parallel Rust workspace tests and repository-wide rustfmt baseline


mblua:
## Context

Issue #1329 cannot satisfy AC13 because the baseline at `4f69e6b3a149ba13175b2da5f48da596992a865e` has five repository-wide rustfmt failures and the ordinary parallel workspace run has shown timing-sensitive shared-state collisions in `cli_test_reset`. Exact reruns at the A cut may pass depending on scheduling; no green run disproves the collision and no later-head failure may be relabelled as A evidence.

This prerequisite restores deterministic isolation and the baseline gates independently before the remaining #1329 rework is approved.

## Scope

- Branch from `4f69e6b3a149ba13175b2da5f48da596992a865e`.
- Fix only the `cli_test_reset` cross-process shared-state collision.
- Add deterministic Windows holder/contender regression coverage for acquisition, bounded contention/timeout, owner abandonment, release, and successful retry.
- Apply rustfmt only to these baseline files:
  - `src-tauri/src/commands/session.rs`
  - `src-tauri/src/pty/local_backend.rs`
  - `src-tauri/src/web/commands.rs`
  - `src-tauri/tests/pty_lifecycle_regression.rs`
  - `src-tauri/tests/pty_powershell_managed_native.rs`
- Do not include any workgroup-event feature code from #1329.

## Acceptance criteria

- The reset lock uses a finite documented wait and explicitly handles `WAIT_OBJECT_0`, `WAIT_ABANDONED`, `WAIT_TIMEOUT`, `WAIT_FAILED`, and unexpected results.
- Win32 creation/wait/release/close errors retain the primary error, report secondary cleanup failures safely, and preserve the established JSON command-error contract.
- Every unsafe block has an accurate `// SAFETY:` proof, including same-thread mutex ownership/drop.
- Deterministic Windows holder/contender tests cover wait, timeout, abandonment, release, and retry.
- `cargo test --manifest-path src-tauri/Cargo.toml --workspace` passes with ordinary parallel execution.
- `cargo fmt --manifest-path src-tauri/Cargo.toml --all -- --check` passes.
- `cargo check --manifest-path src-tauri/Cargo.toml --workspace` passes.
- `cargo clippy --manifest-path src-tauri/Cargo.toml --workspace --all-targets -- -D warnings` passes.
- A/B evidence records:
[...]

PR #1402, opened 2026-08-17
fix: register the screenshot global hotkey at the top of setup (#1398)

### LINKED ISSUE #1398: fix: screenshot hotkey dead during startup restore and titlebar chip never appears (regression from a308271c / #1341)
## Situation

The screenshot global hotkey (`screenshotCaptureHotkey`, e.g. `Ctrl+2`) does not respond during application startup, and the titlebar chip that displays the configured shortcut never appears at all.

This is a **functional regression**, not a cosmetic one. The dead-hotkey window spans from window startup to the end of the post-restore startup tail. On a light instance it measured ~600 ms, but it **scales with the number of sessions to restore**, so an instance with many sessions waking on startup can leave the capture hotkey unresponsive for several seconds.

## Cause

Commit `a308271c` ("fix(1341): run startup restore off the main thread so the auto-update prompt can render", PR #1344, refs #1341) moved the startup restore out of a main-thread `block_on` into a spawned runtime task. That change was correct for its own goal, but the hotkey registration block was physically the last statement of `.setup(...)`, adjacent to the post-restore service-start block, and was carried along with the moved range. It now lives as the **last line of the post-restore tail** (`src-tauri/src/lib.rs:1797-1807`).

Two consequences:

1. The registration no longer completes before the WebView loads — it now races with it, and runs only after the full restore body and every startup service (observers, resource monitor, reaper, mailbox poller, loop scheduler, auto-close, non-stop watchdog, ui-automation).
2. `ScreenshotHotkeyStatusChip` (`src/sidebar/components/Titlebar.tsx:37-82`) reads `screenshot_get_hotkey_status` exactly once in `onMount`, with no retry, polling, or event listener. It observes the default `ScreenshotHotkeyRuntime` (`registered: false`) and hides itself permanently for that session.

The verbatim-move audit performed on PR #1344 compared the moved code against itself; it did not evaluate the change of startup phase, which is the axis that broke. Neither #1341, PR #1344, its (empty) reviews, nor the closing comment mentions the screenshot hotkey beyond listing it among the moved items.

`register_configured_hotkey` has **no dependency on the restore**. Its three requirements are all managed before the first line of `.setup(...)`: the `tauri_plugin_global_shortcut` plugin (registered in the builder), `ScreenshotHotkeyState`, and `SettingsState`. It is also invoked outside any restore context from hot-path commands (`commands/config.rs`, `commands/screenshot.rs`), confirming it needs only the app handle plus managed state.

## Change

1. **Backend** — `src-tauri/src/lib.rs`: move the hotkey registration block out of the post-restore tail so it no longer waits on the restore. It must not reintroduce a main-thread `block_on` that could starve the WebView, which is exactly what #1341 removed.
2. **Frontend hardening** — `src/sidebar/components/Titlebar.tsx`: replace the single `onMount` fetch with a bounded retry (cancelled on cleanup) so the chip does not depend on winning a startup race.
3. **Tests** — `src/sidebar/components/Titlebar.screenshot-hotkey.test.tsx`: cover the case where the first status response reports `registered: false` and a later one reports the registered hotkey.

## Acceptance

- The configured screenshot hotkey responds **during** the startup restore window, not only after it completes. This is the primary criterion — the chip was the symptom, the dead hotkey is the defect.
- The titlebar chip displays the configured shortcut on a normal boot.

mblua:
Closes #1398

## What was broken

The screenshot global hotkey was dead during startup and the titlebar chip showing the configured shortcut never appeared.

Commit `a308271c` (PR #1344, refs #1341) moved the startup restore off the main thread so the auto-update prompt could render. That was correct for its own goal, but the hotkey registration block was physically the last statement of `.setup(...)`, adjacent to the moved service-start block, and was carried along into the post-restore tail. Nothing in #1341, PR #1344, its (empty) reviews, or its closing comment considered the effect — the verbatim-move audit compared the moved code against itself, which cannot catch a change of startup phase.

Result: registration raced the WebView load instead of preceding it. `ScreenshotHotkeyStatusChip` read `screenshot_get_hotkey_status` exactly once in `onMount`, saw `registered: false`, and hid itself permanently for the session. The hotkey itself was unresponsive for the whole window between window startup and the end of the tail — a window that scales with the number of sessions being restored.

## What changed

- **`src-tauri/src/lib.rs`** — the registration block moves out of the post-restore tail into its own `tauri::async_runtime::spawn` task at the top of `.setup(...)`. No main-thread `block_on` is reintroduced (that is what #1341 removed); the `SettingsState` read is `.read().await` inside the spawned task. It mirrors the existing #1327 agent-update task, so it is not a new pattern. Adds one `log::info!` on success; the failure-path text is unchanged so existing log scraping stays valid.
- **`src/sidebar/components/Titlebar.tsx`** — the one-shot `onMount` fetch becomes a bounded re-read (20 attempts at 250 ms, cancelled on cleanup), so the chip no longer depends on winning a startup race.
- **`src/sidebar/components/Titlebar.screenshot-hotkey.test.tsx`** — adds the race case: first status reports `registered: false`, a later one reports the registration.

`src-tauri/module-arcs.txt` is untouched: zero module arcs added or removed, SCC member sets identical.

## Verification

| Gate | Result |
|---|---|
| `cargo clippy --workspace --all-targets -- -D warnings` | clean, exit 0 (re-run independently by grinch) |
| `cargo test` | 3733 passed, 0 failed, 28 ignored |
| `npm run typecheck` | exit 0 |
| `npm test` | 1601 passed across 157 files |
| module-arcs diff | empty |

Acceptance criteria, all measured on an isolated test instance with its own config dir:

- **Ordering** — `[screenshot] global hotkey registered '<key>'` precedes `[restore] complete` in `app.log`, reproduced across every boot. At the base the ordering is structurally reversed, so observing the forward order at all falsifies the base.
- **Hotkey live during restore** — with 30 sessions restoring, a real `keybd_event` chord was serviced **618 ms before** `[restore] complete` (press outcome at log line 77, `[restore] complete` at line 204, with 30 `[restore] woke` lines still to come). Reproduced at N=12. The outcome was `capture failed before overlay: selectionCoordinatorBusy` — the accepted failure surface, proving the hotkey is live rather than silent.
- **Chip** — renders the configured shortcut on a normal boot.

Adversarial review: PASS with zero defects. Each frozen plan block was compared byte-for-byte against the committed region. Both risk claims were attacked and held: an early press cannot panic or hang (`begin_capture_from_hotkey` spawns its own task, every reachable `app.state::<T>()` is managed before `.setup(...)`, and `surface_hotkey_failure` is panic-free by construction), and double registration is closed more strongly than claimed — `app.global_shortcut()` has exactly one call site in the repo, and the `ScreenshotHotkeyState` mutex is held across check-and-register, so it is atomic rather than TOCTOU-prone.

Two review findings are recorded on the issue rather than fixed here, since this change is scoped out of `src-tauri/src/screenshot/`: a failure event dropped in the narrow window before the frontend attaches its listener (strictly no worse than the base, where the press did nothing at all), and a new coupling where the state mutex is held across a main-thread round-trip — not reachable today, but worth knowing. The issue also records that the hotkey actually becomes live on the first event-loop drain rather than at the top of `setup`, and that a multi-session instance is required if the ordering check is ever reused as a regression gate.

## Note for a future formatting pass

The inserted Rust block is byte-exact from the certified plan and deliberately not `cargo fmt`-ed; CI has no `cargo fmt --check` step. `rustfmt` would reflow the `match register_configured_hotkey(...)` head and the `log::info!` argument list, so prefer a separate formatting-only commit over folding it in here.

### DIFF
--- a/plans/1398-screenshot-hotkey-startup-race.md
+++ b/plans/1398-screenshot-hotkey-startup-race.md
@@ -1,0 +1,331 @@
+# Plan #1398: Register the screenshot global hotkey at the top of setup instead of the post-restore tail
+
+Author: architect, wg-14. Authored and certified in a single Lite pass on 2026-08-17 UTC.
+
+Status: READY_FOR_IMPLEMENTATION
+
+Issue: [mblua/AgentsCommander#1398](https://github.com/mblua/AgentsCommander/issues/1398), `fix: screenshot hotkey dead during startup restore and titlebar chip never appears (regression from a308271c / #1341)`.
+
+This is a Lite regression fix. It relocates one existing statement block inside `src-tauri/src/lib.rs`, hardens one existing SolidJS component, and adds one test case. It introduces no new production abstraction, no new crate, no new npm dependency, no new module, no new schema, no new Tauri command, no new IPC surface, no new event, no new configuration key and no migration. It adds zero module-to-module dependency arcs.
+
+## 1. Frozen authority and fail-closed entry gate
+
+The implementation working tree is `repo-AgentsCommander`, branch `fix/1398-screenshot-hotkey-startup-race`, targeting `main`.
+
+After `git fetch origin main` on 2026-08-17 UTC, all of the following resolved exactly to `8f272a76b53a15c3c442568ca169e1bb9b7d24cc`:
+
+- committed `HEAD`;
+- `origin/main`; and
+- `git merge-base HEAD origin/main`.
+
+The index and the non-ignored working tree were clean, verified by an empty `git status --porcelain=v1 --untracked-files=all`.
+
+Root `.gitignore` line 11 ignores `/plans/`, so the implementation must force-add this exact plan file with `git add -f plans/1398-screenshot-hotkey-startup-race.md`. Do not remove or weaken the repository's `plans/` ignore rule.
+
+Immediately before implementation, fetch `origin/main` again. Stop for current-base review if `origin/main`, the local committed branch head, or their merge base no longer equals the frozen SHA above. Do not rebase, merge a moved base, or silently substitute a newer commit under this certification.
+
+Branch-name validation was checked against `scripts/validate-branch-name.mjs` line 15. `fix/1398-screenshot-hotkey-startup-race` matches the allowed pattern with type `fix`, number `1398`, and slug `screenshot-hotkey-startup-race` (29 characters, under the 50-character cap), so the required `validate-branch-name` check will pass.
+
+Every line number in this plan refers to the frozen SHA above. If a line number no longer matches the quoted text, stop and re-anchor on the quoted text, never on the number.
+
+## 2. Objective and non-goals
+
+Objective: make the configured screenshot global hotkey live from the very beginning of `.setup(...)` instead of after the whole startup restore plus every startup service, and make the titlebar chip stop depending on winning that race.
+
+Non-goals, binding on the implementer:
+
+- Do not reintroduce a main-thread `tauri::async_runtime::block_on` for the settings read. `src-tauri/src/lib.rs:1034-1040` declares the spawned-task launch as the only sanctioned way to start the startup restore precisely because a main-thread `block_on` starves WebView2 while a session open awaits the #1327 update gate. This plan never touches `spawn_restore_startup`'s launch shape.
+- Do not change `register_configured_hotkey`, `parse_screenshot_hotkey`, `begin_capture_from_hotkey`, or anything else under `src-tauri/src/screenshot/`.
+- Do not change `src/shared/ipc.ts`, `src/shared/types.ts`, or any Tauri command. See section 7.
+- Do not add a backend event, a push notification, or a hotkey-registration broadcast. The frontend hardening is a bounded re-read of the existing `screenshot_get_hotkey_status` command.
+- Do not reorder, rename, or restructure any other statement in `.setup(...)` or in the post-restore tail.
+- Do not add tests beyond the single frontend case in section 6.
+
+## 3. Verified current state
+
+Verified by direct read at the frozen SHA:
+
+- `src-tauri/src/lib.rs:1797-1807` holds the hotkey registration as the last statement of the post-restore tail inside `spawn_restore_startup`. The tail begins at `lib.rs:1749` and runs only after the restore body completes; `[restore] complete` is logged at `lib.rs:1700-1701`, before `completion.complete()` at `lib.rs:1744` and therefore before the tail. Today the hotkey is registered strictly after the `[restore] complete` line reaches the log.
+- `.setup(move |app| {` begins at `lib.rs:2110`. Everything `register_configured_hotkey` needs is already in place before that line: the `tauri_plugin_global_shortcut` plugin at `lib.rs:2062-2073`, `.manage(settings)` (type `SettingsState`, built at `lib.rs:1991`) at `lib.rs:2084`, and `.manage(screenshot_hotkey_state)` at `lib.rs:2107`.
+- `lib.rs:45` already has `use tauri::{Emitter, Manager};` and `lib.rs:32` already has `use config::settings::SettingsState;`, so the relocated block needs no new import.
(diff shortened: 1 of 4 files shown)


mblua:
## Merged with `--admin` over a red check — why, and what the red check is

The `rust-regression` (Windows) job fails on `cli::task_ops::tests::concurrent_writes_return_correct_post_edit_content`. That failure is **unrelated to this change** and predates it. Merging over it is a deliberate, user-authorized decision, recorded here rather than left implicit.

### The failure

```
thread '...concurrent_writes_return_correct_post_edit_content' panicked at src-tauri\src\cli\task_ops.rs:1418:26:
h1 unexpected outcome: Err(LockIo("...\wg-1\TASK.md.lock", Os { code: 5, kind: PermissionDenied, message: "Access is denied." }))
```

`create_new` on `TASK.md.lock` returned Win32 code 5 (`PermissionDenied`). `LockGuard::acquire` (`src-tauri/src/cli/task_ops.rs:427-471`) retries **only** `ErrorKind::AlreadyExists` and returns `LockIo` for anything else; the test's `match` has no arm for `LockIo`, so it falls through to `other =>` and panics. That proximal chain is certain. **The ultimate cause — why the runner's kernel returned code 5 — is not established.**

### Why this change cannot be responsible

Four independent reasons, any one sufficient:

1. **`src-tauri/src/cli/task_ops.rs` is byte-identical between base and head** — blob `ef4e3913417a40a0dd6d9547f51e5352479446e9` at both `8f272a76` and `2898161`. This PR touches only `src-tauri/src/lib.rs`, two frontend files, and its plan.
2. **The test uses no tokio.** It is a plain `#[test]` over `std::thread::spawn` and `std::sync::Barrier`, so it cannot be affected by runtime worker-thread availability.
3. **The unit-test binary never builds the production `.setup(...)`.** The shared helper is `tauri::Builder::default().any_thread()` with no `.setup(...)`, so the code this PR changed does not execute in the test binary at all.
4. **The same commit passed the same job.** The `push`-event run (job `95432334588`) reports this test `... ok` with `3485 passed; 0 failed`, on the **same runner image and version** (`windows-2025-vs2026`, `20260810.198.2`) as the two failing runs. Non-determinism is therefore established at the environment level, not inferred.

Three data points on this commit: `push` passes, PR attempt 1 fails, PR attempt 2 fails. The panicking arm moved between attempts (`h2` at line 1423, then `h1` at line 1418), which is the signature of a symmetric race between the two writers rather than a deterministic ordering bug.

### What was ruled out, and honestly retracted

An initial diagnosis attributed the code 5 to Windows delete-pending semantics. **That was tested and retracted**: on a local Windows box, deleting a lock file with a live handle yields code 80 (`ERROR_FILE_EXISTS` → `ErrorKind::AlreadyExists`), which `acquire` already retries. Modern Windows unlinks the name immediately (POSIX delete semantics), so the proposed window never opened. Code 5 could not be reproduced in any scenario.

Consequently the "obvious" three-line fix — making `acquire` retry code 5, mirroring what `do_rename` already does ten lines below for transient AV holds — was **withdrawn**, because it would suppress an unidentified condition rather than fix a known one.

Local reproduction attempts: 2900 isolated races and 5 full `cargo test --lib` runs, all green. Explained by `RealTimeProtectionEnabled = False` on that machine versus Defender active on GitHub runners; the CI runner is also 5-7x slower on this suite, implying far more filesystem pressure.

### Follow-up

The correct next step is to instrument `acquire` to log `raw_os_error`, path state, and attempt number, then run it on the Windows runner and fix from what that shows. That belongs to **#1343** ("Restore deterministic parallel Rust workspace tests..."), which is already open for Rust test non-determinism.

### This PR's own verification

All green: `cargo clippy --workspace --all-targets -- -D warnings` clean (re-run independently during review), `cargo test` 3733 passed / 0 failed, `npm run typecheck` exit 0, `npm test` 1601 passed, `module-arcs.txt` untouched, plus the three measured acceptance criteria in the PR body — including a hotkey press serviced 618 ms before `[restore] complete` with 30 sessions restoring.

Issue #1261, opened 2026-08-06
flaky: capped_shutdown_aborts_a_held_worker_then_cleans_pending_resources_and_keys fails under CPU contention


mblua:
## Situation

`session::selection::tests::capped_shutdown_aborts_a_held_worker_then_cleans_pending_resources_and_keys`
(`src-tauri/src/session/selection.rs:5357`) fails intermittently under CPU contention:

```
assertion failed: left: 0, right: 1
```

Observed during the merge-resolution review for #1252.

## Evidence

- The test runs against a **25 ms deadline** (`budget=25ms finalizationReserve=6.25ms`) and every log
  line at failure time was a deadline expiry (`retained after absolute deadline`).
- It failed while `tsc` was competing for CPU on the same machine.
- On a clean run afterwards it **passed**; isolated it passes **5 of 5** in 0.03–0.05 s.

## Why it cannot be #1252

Measured by transitive reachability over `src-tauri/module-arcs.txt`: `session::selection` reaches 126
modules, and **none of them** is `loops::events`, `loops::scheduler` or `commands::loops`.

It reaches `config::project_settings` only via a 4-hop path
(`selection → agentscommander_lib → commands::ac_discovery → web::commands → config::project_settings`),
and the only change #1260 made to that module is adding a `bool` data field — no locks, no tasks, no
timing.

## Not yet diagnosed

The correlation with CPU contention and a 25 ms deadline is what was observed; it is **not a confirmed
root cause**. Confirming it requires reproducing under controlled load with enough logging to show
what missed the deadline.

## Distinct from the other five

Sixth intermittent test recorded during #1252 verification, each in a different place with a different
mechanism: #1253 (`pty::container_backend`), #1254 (`session::selection`, two other tests), #1255
(`session::context_alerts`), #1256 (`commands::ac_discovery`), #1258 (`pty::git_watcher`).

This one is in `session::selection` like #1254, but a different test with a different assertion — the
25 ms shutdown deadline rather than `critical_keys` bookkeeping. Confirm they share a cause before
merging them.

Issue #1258, opened 2026-08-06
flaky: detect_git_status_stays_red_from_modify_through_commit_until_push fails under full-suite load


mblua:
## Situation

`pty::git_watcher::tests::detect_git_status_stays_red_from_modify_through_commit_until_push`
(`src-tauri/src/pty/git_watcher.rs:827`) fails intermittently under full-suite load:

```
panicked at src-tauri\src\pty\git_watcher.rs:827:9: synchronized baseline
```

Observed during the final verification work for #1252.

## Evidence

- **Isolated: 10 runs, 10 green.**
- **Full suite: failed once, then 2 consecutive green runs** on the same tree — intermittent.
- **The branch that observed it does not touch `src-tauri/src/pty/` at all.**

The assertion is `synchronized baseline`, i.e. synchronization with a git subprocess. Same
environmental family as #1256 (subprocess spawn under parallel load), but a different test in a
different module.

**That is the observed correlation, not a confirmed root cause.** Confirming it needs the subprocess
outcome captured at failure time and a reproduction under load.

## Distinct from the other four

- **#1253** `pty::container_backend` — shared shutdown lock deadline
- **#1254** `session::selection` — `critical_keys` bookkeeping (two tests)
- **#1255** `session::context_alerts` — liveness-check timeout
- **#1256** `commands::ac_discovery` — subprocess spawn failure
- **this one** `pty::git_watcher` — git subprocess synchronization

Note this is **not** #1253: that one is `pty::container_backend`, a different module and a different
mechanism.

## Context

This is the fifth intermittent test recorded during #1252 verification, each in a different module
with a different mechanism. None was introduced by that branch. If suite-level flakiness is ever taken
on as one workstream, #1253 through #1257 are the known inventory as of `8da11a66`.

Issue #1256, opened 2026-08-06
flaky: stage_e_later_user_rule_and_excludes_win_over_managed_negation fails under full-suite load


mblua:
## Situation

`commands::ac_discovery::tests::stage_e_later_user_rule_and_excludes_win_over_managed_negation`
(`src-tauri/src/commands/ac_discovery.rs:4636`) fails intermittently under full-suite load:

```
panicked at src-tauri\src\commands\ac_discovery.rs:4636:13:
assertion failed: init.success()
```

Observed during the round-3 review work for #1252.

## Evidence

- **Isolated: 10 runs, 10 green.**
- **Full suite on the same tree: 4 green, 1 red**, with the last three consecutive runs green.
- **The branch that observed it does not touch `src-tauri/src/commands/ac_discovery.rs`.**

The assertion is `init.success()` — a subprocess spawn. Spawn failures under parallel load on Windows
are a known shape of environmental flakiness, but **that is the visible correlation, not a confirmed
root cause.** Confirming it requires capturing the child's stderr and exit status at failure time and
reproducing under load.

## Distinct from #1253, #1254 and #1255

Four intermittent tests have now been observed, each with a different mechanism:

- **#1253** `pty::container_backend` — deadline on a shared shutdown lock
- **#1254** `session::selection` — `critical_keys` bookkeeping not yet cleared (two tests)
- **#1255** `session::context_alerts` — liveness-check timeout
- **this one** `commands::ac_discovery` — subprocess spawn failure

They share a family — assertions coupled to timing or to shared resources under parallel load — but
they are four different mechanisms in four different modules. Do not merge them without evidence of a
common cause.

If someone wants to treat suite-level flakiness as one workstream, these four are the known
inventory as of `7b8dee8e`.

Issue #1255, opened 2026-08-06
flaky: one_global_delivery_slot_keeps_samples_and_maintenance_live times out under full-suite load


mblua:
## Situation

`session::context_alerts::one_global_delivery_slot_keeps_samples_and_maintenance_live` fails under
full-suite load with:

```
timed out waiting for scripted liveness checks
```

Observed during the round-3 review of #1252, on commit `71831b47`.

## Evidence

- Passes in isolation on an unloaded machine.
- Fails only under the load of the full `cargo test --lib --bins --tests` run.
- The branch that observed it does not touch `src-tauri/src/session/` at all.

Preexisting timing-sensitive test, not a regression.

## Distinct from #1253 and #1254

- **#1253** is `pty::container_backend::...poisoned_global_sweep...` — different module, deadline on a
  shared shutdown lock.
- **#1254** is in `session::selection` and turns on `critical_keys` bookkeeping.
- **This one** is in `session::context_alerts` and is a liveness-check timeout.

Same family of problem — assertions coupled to wall-clock behaviour under parallel load — but three
different tests with three different mechanisms. Do not merge them without evidence they share a
cause.

## Not yet diagnosed

The above is what was observed, not a root cause. Confirming it requires reproducing under load with
enough logging to show what starved the liveness checks.

PR #1465, opened 2026-08-20
feat(#1462): record native-tool usage in the RTK savings database

### LINKED ISSUE #1462: Record Read/Grep/Glob/Edit/Write tool usage in the RTK savings DB via Claude Code hooks
## Situation

AgentsCommander seeds a Claude Code `PreToolUse` hook (`.ac/default.claude/hooks/ac_rtk_claude_Bash.js`, `ac_rtk_claude_PowerShell.js`, `ac_rtk_shared.js`) into every workgroup replica. It rewrites shell commands to their `rtk` equivalents, so `rtk` records them (tokens, savings, timing) in the SQLite DB named by `RTK_DB_PATH`. Verified on 2026-08-20 in the wg-4 tech-lead replica: `cat X` through the Bash tool becomes `rtk read X` and lands as a row in the `commands` table; `Get-Content X` through the PowerShell tool is passed through unchanged and appended to the ignored-tools log.

Claude Code's native file tools never reach `rtk`: `Read`, `Grep`, `Glob`, `Edit` and `Write` run in-process, no hook is registered for them (`settings.local.json` only has `Bash` and `PowerShell` matchers), and after a `Read` call both the DB row count and the ignored log were unchanged. The largest share of what an agent actually reads is therefore invisible in `rtk gain`, `rtk gain -H` and `rtk gain -f json|csv`.

## Why

Operators use the RTK DB as the record of what each agent consumed. Without the native tools the numbers are partial: token volume, which files are read or re-read, and how often, cannot be seen. Filtering the content through `rtk` is not possible (native tools cannot be proxied, and denying `Read` to force `cat` would break `Edit`, which requires a prior `Read` of the file), but recording the usage is.

## How

Add an observability-only hook for the native tools, mirroring the existing Bash/PowerShell hooks:

- A new hook script under `hooks/` (final name fixed by the plan) registered in `settings.local.json` for both `PreToolUse` and `PostToolUse` with the matcher `Read|Grep|Glob|Edit|Write`.
- `PreToolUse` stores the start time keyed by `tool_use_id`; `PostToolUse` computes `exec_time_ms`, estimates tokens from `tool_input` / `tool_response`, and inserts one row into the existing `commands` table of `RTK_DB_PATH` using Node's built-in `node:sqlite` (Node 24 on the box, no new dependency): `original_cmd` = `<Tool> <file or pattern>`, `rtk_cmd` = a distinctive non-rtk label (for example `tool:read`), `saved_tokens` = 0, `savings_pct` = 0, `project_path` = cwd.
- The hook must never block, slow down noticeably, or alter the tool call: any failure (DB locked, schema mismatch, missing env var, unexpected payload) is swallowed and at most logged.
- Already verified: a synthetic row with `rtk_cmd = 'read-tool'` inserted into a copy of the DB is accepted and listed by `rtk gain` 0.42.4 under "By Command" (`read-tool  1  0  0.0%  63ms`) and counted in the totals.
- Sync the repo mirror `docs/integrations/rtk_claude/` (hooks, `settings.local.json`, README) with the current seed. The seed already split the hook into `_Bash.js`, `_PowerShell.js` and `_shared.js` and added `statusline.sh`; the mirror still holds the old single `ac_rtk_claude.js`. The seed itself lives outside the repo and is updated by the operator from the mirror.

## Caveats

- Rows with zero savings lower the global percentage shown by `rtk gain`; that is the intended visibility, not a regression.
- `exec_time_ms` includes the hook process start-up (about 50 ms per hook), so it measures volume and frequency more than latency.
- The `commands` schema belongs to `rtk` (0.42.4); a future `rtk` migration may require adjusting the insert.

mblua:
Claude Code's native file tools never reached `rtk`, so the largest share of what an agent actually reads was invisible in the savings database. This adds an observability-only hook that records them.

Closes #1462

## What changed

A new `PreToolUse` plus `PostToolUse` hook, `ac_rtk_claude_Tools.js`, matching `Read|Grep|Glob|Edit|Write|NotebookEdit`. `PreToolUse` leaves a timing mark keyed by `tool_use_id`; `PostToolUse` inserts one row into the existing `commands` table of the database named by `RTK_DB_PATH`, using Node's built-in `node:sqlite`. No new dependency. Rows carry `rtk_cmd` labels `tool:read`, `tool:grep`, `tool:glob`, `tool:edit`, `tool:write`, `tool:notebookedit`, and `rtk gain` lists them beside real `rtk` invocations.

The hook cannot filter what a native tool returns; native tools run in-process and are not proxyable. Denying `Read` to force `cat` was rejected because `Edit` requires a prior `Read`. This records usage, it does not save tokens.

Files: `hooks/ac_rtk_claude_Tools.js` and `statusline.sh` created; `settings.local.json`, `ac_rtk_shared.js`, `ac_rtk_claude_Bash.js`, `ac_rtk_claude_PowerShell.js`, `README.md` and `docs/integrations/rtk.md` modified. Nothing deleted. No file under `src-tauri/`, `src/`, `scripts/` or `.github/` is touched.

## Verification

Plan `plans/1462-rtk-native-tools-hook.md`, certified `READY_FOR_IMPLEMENTATION`, `Plan-SHA256` `5B72B6FD16EB0050E0B285DAE73667324CE6C8CD5672430F9E70373B1085BD23`, verified against the git blob by three parties independently.

The committed hook is **byte-identical** to the plan's own section 5.1 literal source, SHA-256 `dc08e4bf35345241681393d7ea669bbb4a470161d83a9336991d58955f2c6983`, so zero interpretation happened between specification and code.

Behavioural suite: 33 of 33 pass, run against the hook extracted from the committed blob and a scratch database, never the live one. One row per tool with the expected fields; eight failure paths (no `RTK_DB_PATH`, missing file, non-database, missing `commands` table, garbage stdin, empty stdin, unlisted tool, `PostToolUseFailure`) each give exit 0, empty stdout and no row; a locked database gives up inside the busy timeout; eight parallel inserts all land; stale marks are swept. A missing database is never created. Static criteria 1 to 17 pass. Reviewed independently, verdict **PASS**, no must-fix defects, and the reviewer re-ran a sample of the behavioural checks rather than accepting the claim.

Zero module arcs added: `ac_rtk_claude_Tools.js` is an isolated node requiring only `node:` builtins.

## Two plan defects, resolved in review without re-certification

Both are defects in the plan's verification instruments, not in the code. The artifact matches the frozen plan byte for byte, so re-certifying would have changed zero shipped bytes.

**Criterion 6 could never pass as written.** It greps `ac_rtk_claude_Tools.js` for `stdout\|logIgnored\|ignored` and requires nothing back, but the plan's own mandated header comment contains "empty stdout" and "ignored-tools log". The implementer reported the contradiction rather than editing the mandated comment to satisfy a broken check. Corrected instrument, use this from now on: `grep -n "process\.stdout\|logIgnored\|ignoredLogPath\|ac_rtk_shared" <hook>` must return nothing, and `grep -n "process\.std" <hook>` must return exactly three lines (`process.stdin` at `:170` and `:171`, `process.stderr.write` at `:179`).

**Criterion 9's grep is a constraint on the README's prose, not only on the code.** `git grep -n "rtk_ignored_tools" HEAD -- docs/integrations/rtk_claude/` must return nothing. The README sentence that records the deliberate divergence from the seed must therefore **describe** the seed's old name rather than **quote** it. Rewording it to name `rtk_ignored_tools.md` breaks criterion 9, which is why it says "the older underscore spelling". Do not "improve" that sentence by naming the file.

## The ignored-log rename, stated accurately

The plan's section 4.6 claimed running replicas already used the new name. That was wrong, and so was the implementer's opposite claim that nothing did. The measured truth, replacing that sentence:

> The operator is renaming the ignored-tools log from `rtk_ignored_tools.md` to `rtk-ignored-tools-claude.md`, so that each coding agent writes its own log; a sibling `rtk-ignored-tools-pi.md` already exists in the Matrix for the non-Claude agent. **The rename has not landed anywhere that survives a spawn.** Measured across the workspace on 2026-08-20: of the 53 replicas on disk, exactly one has hand-edited hooks naming the new file, and its Matrix log is live at 15306 bytes. Every other replica, and the seed at `<workspace>/.ac/default.claude/hooks/ac_rtk_shared.js:58`, still names `rtk_ignored_tools.md`, and ten Matrices hold a live log under the old name. That one hand edit does not survive: `perform_config_seed_with_clock_and_hooks` rebuilds `.claude/` from the seed on every spawn, so the edited replica reverts to the old name the next time it starts. The mirror in this directory therefore carries the new name deliberately and is ahead of the seed on purpose, and the rename becomes real only when the operator applies the copy step of section 8.5. Until then, both spellings are simultaneously live in the workspace, and a reader of a Matrix must check both.

## Operator step, required for any of this to take effect

The files that take effect live in the workspace seed, outside every repository and outside every agent's write zone. **No agent may write them.** The operator copies, from the landed branch to the workspace tree, replacing what is there:

```text
docs/integrations/rtk_claude/hooks/ac_rtk_claude_Tools.js       ->  <workspace>/.ac/default.claude/hooks/ac_rtk_claude_Tools.js
docs/integrations/rtk_claude/hooks/ac_rtk_shared.js             ->  <workspace>/.ac/default.claude/hooks/ac_rtk_shared.js
docs/integrations/rtk_claude/hooks/ac_rtk_claude_Bash.js        ->  <workspace>/.ac/default.claude/hooks/ac_rtk_claude_Bash.js
docs/integrations/rtk_claude/hooks/ac_rtk_claude_PowerShell.js  ->  <workspace>/.ac/default.claude/hooks/ac_rtk_claude_PowerShell.js
docs/integrations/rtk_claude/settings.local.json                ->  <workspace>/.ac/default.claude/settings.local.json
docs/integrations/rtk_claude/README.md                          ->  <workspace>/.ac/default.claude/README.md
```

For this workspace, `<workspace>` is `D:\0_repos\AgentsCommander_iac`. Nothing is deleted from the seed. `statusline.sh` is not copied: the mirror's copy came **from** the seed and is already identical there.

`settings.local.json` hot-reloads, so an already-running session picks up the new registration on its next tool call. The hook **file** must exist before the registration names it, or every matching call silently runs nothing. **Copy the hooks first, `settings.local.json` last.**

## Landed but unproven

Acceptance criteria 22 to 25 are operator-gated and none has been run. Criterion 23 is the one that proves the issue closed. Until an operator applies the copy above and someone reports a live session where a `Read`, a `Grep`, a `Glob`, an `Edit` and a `Write` each add exactly one row and `rtk gain -H` lists them, this change is landed but unproven. The blind spot is not yet demonstrated shut.

## Known limits, deliberate

- Zero-savings rows lower the percentage `rtk gain` reports. That is the intended visibility, not a regression.
- `exec_time_ms` includes the hook's own Node start-up, roughly 40 to 50 ms, so it measures volume and frequency rather than tool latency.
- **Failed tool calls stay invisible.** `PostToolUse` fires only on success; failures raise `PostToolUseFailure`, which is out of scope here.

### DIFF
--- a/docs/integrations/rtk.md
+++ b/docs/integrations/rtk.md
@@ -88,7 +88,7 @@ The wrapped command's own exit code does not change any of this. A command that 
 
 ### The hook covers the filtered set, and nothing else
 
-This section is about RTK's own hook, the one `rtk init` installs. AgentsCommander seeds **different** `PreToolUse` hooks into every workgroup replica, one for the `Bash` tool and one for the `PowerShell` tool, and those also write a log of the commands they declined to rewrite. If you run agents in replicas, read [The AgentsCommander RTK hook for Claude Code](rtk_claude/README.md) before you conclude which hook produced what.
+This section is about RTK's own hook, the one `rtk init` installs. AgentsCommander seeds **different** hooks into every workgroup replica: a `PreToolUse` hook for the `Bash` tool and one for the `PowerShell` tool, which also write a log of the commands they declined to rewrite, and a third registered on both `PreToolUse` and `PostToolUse` for the native file tools, which writes its own rows straight into the database. If you run agents in replicas, read [The AgentsCommander RTK hook for Claude Code](rtk_claude/README.md) before you conclude which hook produced what.
 
 `rtk init` installs both halves of the adoption problem: it writes RTK's instructions into the coding agent's context file, and it patches the agent's configuration with a `PreToolUse` hook that rewrites commands before they run. `--no-patch` skips the patching and prints manual instructions instead. This is the excerpt a patched Claude Code `settings.json` holds under `hooks.PreToolUse`, once per shell tool, not a complete settings file:
 
--- a/docs/integrations/rtk_claude/README.md
+++ b/docs/integrations/rtk_claude/README.md
@@ -1,15 +1,17 @@
 # The AgentsCommander RTK hook for Claude Code
 
-For operators who want to know what a Claude Code agent actually ran, not what it reported. AgentsCommander seeds two `PreToolUse` hooks into every workgroup replica, one for the `Bash` tool and one for the `PowerShell` tool; after this page you can read the two files they produce and say which shell command landed in which one, from which shell, and why.
+For operators who want to know what a Claude Code agent actually ran, not what it reported. AgentsCommander seeds three hooks into every workgroup replica: a `PreToolUse` hook for the `Bash` tool, a `PreToolUse` hook for the `PowerShell` tool, and a third registered on both `PreToolUse` and `PostToolUse` for the six native file tools. After this page you can read the two files they produce and say which shell command landed in which one, from which shell and why, and which native-tool call is recorded in the database.
 
-The hooks, the module they share and their registration are copied into this directory so you can review them without opening a replica:
+The hooks, the module two of them share, their registration and the status line are copied into this directory so you can review them without opening a replica:
 
 - [`hooks/ac_rtk_claude_Bash.js`](hooks/ac_rtk_claude_Bash.js), 83 lines
 - [`hooks/ac_rtk_claude_PowerShell.js`](hooks/ac_rtk_claude_PowerShell.js), 123 lines
-- [`hooks/ac_rtk_shared.js`](hooks/ac_rtk_shared.js), 129 lines, the half that does not depend on a shell, required by both hooks
-- [`settings.local.json`](settings.local.json)
+- [`hooks/ac_rtk_claude_Tools.js`](hooks/ac_rtk_claude_Tools.js), 181 lines, the native-tools hook, which shares no code with the other two
+- [`hooks/ac_rtk_shared.js`](hooks/ac_rtk_shared.js), 129 lines, the half that does not depend on a shell, required by the two shell hooks
+- [`settings.local.json`](settings.local.json), 55 lines
+- [`statusline.sh`](statusline.sh), 17 lines
 
-All four are faithful copies of `<workspace>/.ac/default.claude/`, the seed AC installs from. Do not edit them here: this directory is a mirror, not the source.
+All six are faithful copies of `<workspace>/.ac/default.claude/`, the seed AC installs from, with one deliberate exception: the ignored-log name in this directory is `rtk-ignored-tools-claude.md`, while the seed still carries the older underscore spelling. This directory is ahead on purpose. The operator copies it into the seed, and that copy is how the rename reaches the seed, so do not change the name back to match what the seed says today. Do not edit these files here either: this directory is a mirror, not the source.
 
 ## What the hooks are and where they land
 
@@ -18,11 +20,13 @@ AC seeds `.ac/default.claude/` into each workgroup replica when it creates the r
 ```text
 <workspace>/.ac/<wg-N-name>/__agent_<name>/.claude/hooks/ac_rtk_claude_Bash.js
 <workspace>/.ac/<wg-N-name>/__agent_<name>/.claude/hooks/ac_rtk_claude_PowerShell.js
+<workspace>/.ac/<wg-N-name>/__agent_<name>/.claude/hooks/ac_rtk_claude_Tools.js
 <workspace>/.ac/<wg-N-name>/__agent_<name>/.claude/hooks/ac_rtk_shared.js
 <workspace>/.ac/<wg-N-name>/__agent_<name>/.claude/settings.local.json
+<workspace>/.ac/<wg-N-name>/__agent_<name>/.claude/statusline.sh
 ```
 
-`settings.local.json` registers both hooks, and the whole registration is this:
+`settings.local.json` registers all three hooks across two events, and the whole registration is this:
 
 ```json
 "hooks": {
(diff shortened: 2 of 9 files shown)


mblua:
## Why this is being admin-merged with one red check

`rust-regression` shows one failure on this commit. It is a known flaky test, not a regression, and here is the evidence rather than the assertion.

**The same commit both passed and failed, at the same time.** This repository triggers two parallel workflow runs (push and pull_request) for `23446d08`:

| Run | Job | Result |
| --- | --- | --- |
| [32345741745](https://github.com/mblua/AgentsCommander/actions/runs/32345741745) (push) | [96354015737](https://github.com/mblua/AgentsCommander/actions/runs/32345741745/job/96354015737) | failure |
| [32346210758](https://github.com/mblua/AgentsCommander/actions/runs/32346210758) (pull_request) | [96355394605](https://github.com/mblua/AgentsCommander/actions/runs/32346210758/job/96355394605) | success |

Identical bytes, opposite results. That is the exact signature documented in #1206.

**The failing test cannot be affected by this branch.** The failure is `session::context_alerts::tests::closed_sample_receiver_keeps_due_retry_and_biased_deadline_live`, panicking at `src-tauri/src/session/context_alerts.rs:3699` with "the biased due deadline must run before draining a ready sample backlog". This branch changes nine files, all under `docs/integrations/` and `plans/`. Every Rust blob at this head is byte-identical to `origin/main`, and the incoming `src-tauri/src/pty/output.rs` came verbatim from `e4a97424`. The same test passes on the branch base `1376c2b8`.

**Mechanism:** the assertion races the biased deadline branch against a 1000-message backlog drain and only holds when the scheduler cooperates. Filed as #1466, alongside the existing per-test flake reports #1255, #1256, #1258 and #1261.

Every other required check is green on both runs, including `rust-regression` itself on the pull_request run. The admin merge is for the missing approval (the `main` ruleset requires one and there is no second human reviewer), not to bypass a real failure.

Issue #1598, opened 2026-08-27
Frozen plan digests are not reproducible across checkouts: .gitattributes has no *.md rule


mblua:
## Situation

Frozen delivery plans live at `plans/<stem>.md` and are certified by a SHA-256 digest that architect, dev, and reviewer must all independently reproduce. An implementer is handed that digest cold and told that a mismatch is a blocker and that no design decision may be made without resolving it.

That digest is not reproducible across checkouts today.

`.gitattributes` exists (338 bytes) but contains no rule for `*.md`, so for a plan file:

```
$ git check-attr text eol -- plans/1592-startup-message-surface.md
plans/1592-startup-message-surface.md: text: unspecified
plans/1592-startup-message-surface.md: eol: unspecified
```

With `core.autocrlf=true`, which is the normal Windows clone configuration here, `text` being unspecified means Git treats the file as text and converts LF to CRLF on checkout.

## Measured

On `plans/1592-startup-message-surface.md` at commit `1e9cddfc39d7f48ad8c5acbd9c7e23661fe2d121`:

| Source | Bytes | CRLF | SHA-256 |
| --- | --- | --- | --- |
| Committed blob (`git cat-file blob`) | 18299 | 0 | `96E57EC69BB336A0ABDEE6A822189EDB710A955BE21B1D9C557536B1F25DC164` |
| Current worktree | 18299 | 0 | `96E57EC69BB336A0ABDEE6A822189EDB710A955BE21B1D9C557536B1F25DC164` |
| Fresh checkout (`git checkout-index -f --prefix=...`) | 18478 | 179 | `4897550093E3C2BE4F0D37F4569068B7E2A74BC6FF3583072C36F5059C0D2905` |

179 bytes of difference and an unrelated digest for the same committed content.

## Why this is easy to miss

The blob and the worktree agree **right now**, which makes the problem invisible in the clone where the plan was written. They agree only because the author created the file with LF directly in the worktree and committed it, so no smudge filter has ever run on it.

The divergence appears the first time the file is materialized by Git: a fresh clone, a `git checkout` that recreates it, a worktree add, or CI. That is exactly when an implementer picks the plan up.

## Consequence

[...]

Issue #1591, opened 2026-08-27
feat(portable): preflight the WebView2 runtime with a readable message


mblua:
## Situation

`src-tauri/tauri.conf.json` does not set `webviewInstallMode`, so the NSIS installer falls back to Tauri's default download bootstrapper. A bare portable executable, shipped in a zip with no installer at all, has no bootstrapper and no runtime check whatsoever.

On a Windows machine without the WebView2 runtime (Windows Server, LTSC images, trimmed or offline images), the window cannot be created. The user gets no readable explanation of why the application did not start.

This becomes user-facing the moment we publish a portable zip (#1589), because that path has no installer step that could have installed the runtime.

## Cause

There is no startup preflight for the webview runtime. The failure surfaces only when window creation fails, deep inside Tauri, with no message that names WebView2 or tells the user where to get it.

## How it will be changed

Add a startup preflight that calls `tauri::webview_version()` (returns `Result<String, Error>`) before window creation. On `Err`, print or show one readable message that:

- names the Microsoft Edge WebView2 runtime as the missing dependency,
- says where to obtain it,
- exits through the same readable-error exit path used by the other startup errors, rather than panicking or failing opaquely.

## Out of scope

Embedding a fixed-version WebView2 runtime (roughly 180 MB) in the portable zip. A clear message is preferred over a 180 MB download. Revisit only if demand for fully offline machines appears.

Changing `webviewInstallMode` for the NSIS installer is also out of scope here; the installer path already has a working bootstrapper.

## Acceptance criteria

- On a machine without the WebView2 runtime, launching the executable produces one readable message naming the runtime and where to get it, and a nonzero exit, instead of an opaque failure.
- On a machine with the runtime, the preflight adds no visible behavior and no measurable startup regression.
- The message reaches the user on both the GUI launch path and the console launch path (the console path uses the existing `attach_parent_console` mechanism).

## Context

Requested as item P3 of the Windows portable presentation, coordinated with the portable release-asset work in #1589.

mblua:
## Dependency: the readable message must survive a windowless launch

[...]

Issue #1589, opened 2026-08-27
feat(release): publish a Windows portable zip asset and gate the release on it


mblua:
## Situation

The documentation promises a portable Windows download that the release does not actually produce.

- `docs/quickstart.md` lists, under Windows assets, "`Agents Commander_X.Y.Z_x64-setup.exe` or the portable `agentscommander.exe`".
- `README.md:53` points at GitHub Releases for "the Windows installer, Linux AppImage, macOS dmg, or portable assets".
- `README.md:63` states "Portable: copy the `.exe` to any drive and it carries its own config."

`.github/workflows/release.yml` publishes no such asset. The Windows leg uploads `agentscommander-windows-x86_64.exe`, a raw binary that exists so `npm/install.js` can download and checksum it, plus the NSIS installer. There is no portable download, no accompanying license or notice files, and nothing telling a user where the instance state will land.

A user who follows the docs today therefore downloads the raw npm asset and runs it under its published name. That name breaks instance identity: `binary_suffix()` (`src-tauri/src/config/profile.rs:27`) splits the file stem on the **first** underscore, so `agentscommander-windows-x86_64.exe` yields the suffix `64`. The result is an instance with

- its config directory at `.agentscommander-windows-x86_64\` instead of `.agentscommander\`,
- an `AC [64]` badge in the titlebar,
- a non-default single-instance mutex (`profile.rs:80`), and
- hashed ports 9886 and 9906 instead of the documented 9877 and 9887 (`profile.rs:159` and `:181`).

None of that is a defect in the app: the portable contract in `docs/features/portable-instances.md` is real and works. It is a packaging gap. The binary just needs to ship under its canonical name, with the files a standalone download requires.

## Proposed change

Add a Windows portable zip to the release, built in the existing Windows matrix leg of `release.yml`, and make the release gate refuse to publish without it.

1. **Build the asset.** In the Windows leg, after the existing raw-binary upload, assemble `agentscommander-X.Y.Z-windows-x86_64-portable.zip` containing:
   - `agentscommander.exe` (the binary under its canonical name, so instance identity resolves to prod defaults),
   - `LICENSE`,
   - `THIRD_PARTY_NOTICES.md`,
   - `PORTABLE.txt`: where the instance directory is created, how to run a second independent instance by renaming the exe, that the WebView2 runtime is required, that Windows artifacts are currently unsigned (#717), and how to update in place without losing state.
[...]

Issue #1577, opened 2026-08-27
npm global install fails: instance config dir resolves next to the binary in a root-owned path


mblua:
## Situation

Installing the npm package globally on Linux and then launching the app fails. Reproduced by a user on Ubuntu with `@mblua/agentscommander@0.30.3`.

Without `sudo`, the app panics before the window opens:

```
$ agentscommander
[instance-gitignore] warning: failed to create instance config directory /usr/local/lib/node_modules/@mblua/agentscommander/bin/.agentscommander: Permission denied (os error 13)
[INFO] agentscommander_lib::config::settings - [cli] No settings file found at "/usr/local/lib/node_modules/@mblua/agentscommander/bin/.agentscommander/settings.json", using defaults

thread 'main' (39863) panicked at src-tauri/src/lib.rs:1868:47:
Failed to create app outbox directory: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }
```

With `sudo` it boots, but every artifact (settings, tokens, `daemon.pid`, `instances/`, `app.log`) is written under `/usr/local/lib/node_modules/...` and owned by root, so the app then requires root forever and any npm reinstall of the package destroys the user's state.

## Cause

`resolve_instance_location` in `src-tauri/src/config/mod.rs:87` resolves the config directory as **portable, next to the executable**: `<exe_parent>/.<exe_stem>` (primary branch at `src-tauri/src/config/mod.rs:120`). The `$HOME` fallback at `src-tauri/src/config/mod.rs:149` is reached only when `current_exe()` itself fails, never when the resolved directory is not writable.

A global npm install places the downloaded binary in `<npm_prefix>/lib/node_modules/@mblua/agentscommander/bin/`, which with the default prefix `/usr/local` is root-owned. The portable rule therefore points the entire instance state at a directory the invoking user cannot write.

Two consequences make the failure worse than a warning:

1. `std::fs::create_dir_all(&app_outbox_path).expect("Failed to create app outbox directory")` at `src-tauri/src/lib.rs:1869` aborts the process with a raw panic and a backtrace hint instead of a readable startup error. The earlier `[instance-gitignore]` warning already knew the directory was unwritable and boot continued anyway.
2. There is no supported escape hatch in a release build: `AGENTSCOMMANDER_TEST_CONFIG_DIR` (`src-tauri/src/config/mod.rs:163`) is behind `#[cfg(debug_assertions)]`, so a shipped binary cannot be told to store state elsewhere.

[...]

PR #1599, opened 2026-08-27
fix(portable): surface the second-instance startup message, refs #1592

### LINKED ISSUE #1592: fix(portable): second instance is silent, and fatal startup messages do not survive a windowless launch
## Situation

Two failures with one shape. In both, the user launches the executable by double-click and nothing readable happens.

**A. A second instance exits in silence.**

`src-tauri/src/main.rs:129`:

```rust
// Another GUI instance is already running; exit silently
std::process::exit(0);
```

The single-instance mutex and the web server ports derive from the executable name suffix only (`src-tauri/src/config/profile.rs:85` and `:160`), never from the folder the executable lives in. Two copies of `agentscommander.exe` in different directories therefore share one mutex.

That is exactly the shape the portable presentation creates: a user who already has an installed copy running, then unpacks the portable zip (which ships the binary as plain `agentscommander.exe`, see #1589 and PR #1593) and double-clicks it, gets complete silence. Nothing tells them the launch was refused or why.

Renaming the portable copy to `agentscommander_<name>.exe` already yields a fully independent instance, with its own mutex, config directory, and ports. The user has no way to discover that.

**B. A fatal startup message printed to a console does not survive the exit.**

`src-tauri/src/main.rs:1` sets `windows_subsystem = "windows"` for non-debug builds, so a release binary launched by double-click has no inherited console.

`cli::attach_parent_console()` (`src-tauri/src/cli/mod.rs:247`) handles that: when `AttachConsole(ATTACH_PARENT_PROCESS)` fails it calls `AllocConsole()`, so a console is created. But every caller in `main.rs` follows the `eprintln!` with an immediate `std::process::exit`, and an allocated console is destroyed when the process exits. The user sees a black window appear and vanish.

A message that is readable for 40 milliseconds is not a readable message. Functionally, failure B degrades into failure A: the app was double-clicked and nothing happened.

## Why these are one issue

Failure B is not a separate defect that happens to be nearby. It is the reason failure A's fix does not work if written naively: replacing the silent `exit(0)` with an `eprintln!` plus `exit(1)` produces a message that flashes and disappears, which is the same user experience we started with. A's acceptance criterion is unsatisfiable without B.

Two other issues depend on the same surface and are deliberately kept out of this one:

- **#1577** replaces the panic at `src-tauri/src/lib.rs:1869` with a readable startup error for an unwritable state directory.
- **#1591** adds a WebView2 runtime preflight with a readable message.

Both have exactly this shape: the GUI cannot start, so the only surface left is a console that is about to be destroyed. This issue builds the shared surface and proves it on the smallest case; #1577 and #1591 consume it rather than each inventing one.

## How it will be changed

**1. A shared fatal-startup-message surface.** One helper that guarantees the following outcome:

> When the process was launched without an inherited console, a fatal startup message must remain readable by the person who launched it until they dismiss it or read it.

The mechanism is a plan decision, not fixed here: a native message box, or holding the allocated console open until acknowledged, are both acceptable. It must not regress the case where a console *was* inherited (launched from a terminal), where the current `eprintln!`-to-stderr behavior is correct and must be preserved verbatim so that scripting and CI keep working.

**2. Use it for the second-instance case.** Replace the silent `exit(0)` at `src-tauri/src/main.rs:129` with a message that states:

- an instance of this identity is already running,
- renaming the executable to `agentscommander_<name>.exe` gives an independent instance with its own configuration directory and ports.

The `ui_automation_enabled` branch immediately above keeps its current behavior; only the plain GUI branch changes.

## Out of scope

- Focusing the already-running window through `tauri-plugin-single-instance`. Nicer, larger, not requested.
- The #1577 and #1591 call sites. They consume the surface in their own deliveries.
- Changing what any non-fatal path prints.

## Acceptance criteria

- Launching a second instance of the same identity, by double-click, from a release build, produces a message naming the cause and the rename workaround that **remains readable until the user dismisses or reads it**.
- Launching a second instance from a terminal that already has a console keeps writing to that console's stderr, with unchanged text and exit code.
- The `ui_automation_enabled` branch's behavior and exit codes are unchanged.
- Launching a copy renamed to `agentscommander_<name>.exe` still starts normally as an independent instance.
- The surface is exposed as a reusable helper that #1577 and #1591 can call without modification.

## Context

Item P4 of the Windows portable presentation, plus the cross-cutting readable-error requirement raised against P1 and P3. Coordinated with the portable release-asset work in #1589 / PR #1593.

mblua:
Closes #1592

## What and why

A Windows release binary is built with `windows_subsystem = "windows"`, so a second instance of the same executable identity hit `std::process::exit(0)` and the user saw nothing at all: they double-clicked and the app appeared not to launch.

Printing before that exit is not sufficient. `attach_parent_console()` allocates a process-owned console when no usable handle is inherited, and that console is destroyed when the process exits, so the message would flash for milliseconds and vanish. That is functionally identical to the silence it was meant to replace.

This is user-facing now rather than theoretical: the Windows portable zip (#1589, PR #1593) ships the binary as plain `agentscommander.exe`, so anyone with an installed copy running who unpacks and launches the portable one hits exactly this.

## Changes

- **`src-tauri/src/cli/mod.rs`** — adds `present_fatal_startup_message(&str)`, plus private seams `standard_handle_state`, `standard_handle_is_invalid`, `startup_message_destination`, and `show_native_startup_message`. It shows a blocking `MessageBoxW` **only** when stdout and stderr are both invalid, and otherwise writes to the valid stream exactly as before. `attach_parent_console()` is refactored onto the same shared predicate so the two cannot drift.
- **`src-tauri/src/main.rs`** — the plain second-instance branch now presents the message and retains exit code 0. Both UI-automation branches are unchanged.
- **`scripts/smoke-cli-powershell.ps1`** — two bounded second-instance cases for the plain binary: real pipe capture and child-shell file redirection, each with a 15 s `WaitForExit` and scoped `taskkill /PID <pid> /T /F` cleanup.
- **`plans/1592-startup-message-surface.md`** — the frozen delivery plan.

The helper returns `()` and never exits; the caller owns termination. That is what lets this issue keep exit 0 while #1577 and #1591 later exit nonzero off the same surface.

## The constraint that shaped the design

The predicate is **handle validity**, never console presence. A `GetConsoleWindow() == null` test would be wrong in a way that costs hours: a script launching the binary with `RedirectStandardOutput`/`RedirectStandardError` produces a child with valid redirected handles and no console, so a console-presence predicate would pop a modal on a headless CI runner with nobody to dismiss it and block until the job timed out.

That was measured, not assumed. Independently, by two reviewers, under both Windows PowerShell 5.1 and `pwsh`:

- direct `ProcessStartInfo` capture yields `FILE_TYPE_PIPE` on both handles
- `Start-Process -RedirectStandardOutput/-RedirectStandardError <file>` yields `FILE_TYPE_DISK` on both handles

Both valid, so both select a stream destination. The new redirected-file smoke case is the permanent detector for this: a wrong predicate becomes a bounded, legible test failure instead of a silent CI hang.

The refactor also preserves the guard from **#129** exactly, including the null-first short circuit so `GetFileType` never receives a null handle. Unconditional `AttachConsole` rebinds valid inherited pipe handles and drops captured output; that is why the guard exists.

## Verification

Three-party plan consensus (architect, dev, adversarial reviewer) on plan blob SHA-256 `96E57EC69BB336A0ABDEE6A822189EDB710A955BE21B1D9C557536B1F25DC164`, then implementation review **PASS**, then a focused merge-resolution review **PASS**.

Local checks against merged `main`, all exit 0: `cargo check --locked --all-targets`, `cargo clippy --locked --workspace --all-targets -- -D warnings`, `cargo test --locked --lib --bins --tests`, `npm run build:prod:no-bundle`, `npm run smoke:cli-release-windows` (`passed=4 skipped=0 failed=0`), `git diff --check`.

`cargo fmt --all -- --check` exits 1. Proven pre-existing: an untouched checkout of merged base `95db68c5` and this head produce byte-identical formatter output, 266 diff headers across 30 files, none of them a file this PR touches. No CI workflow runs `cargo fmt`.

Dependency/layering gate against the merged base: 191 modules, 1,037 unique arcs, arc-set symmetric difference **0**, the one pre-existing 85-member cyclic SCC identical before and after, and `src-tauri/module-arcs.txt` byte-identical.

Mutex collision in the new smoke was verified by negative probe: pre-holding `Local\AgentsCommander_SingleInstance` makes the harness fail rather than pass, proving the cases cannot pass while testing nothing.

## Not covered here

Three checks are release-owner manual gates and are deliberately not claimed as automated: the Explorer double-click dialog, a real `FILE_TYPE_CHAR` terminal, and the renamed `agentscommander_<name>.exe` instance. The automated set cannot cover a native dialog or a real console, and saying so beats claiming coverage that does not exist.

Follow-ups filed: #1598 (frozen plan digests are not reproducible across checkouts; `.gitattributes` has no `*.md` rule). #1577 and #1591 consume this surface next.


mblua:
CI attribution note, recorded here because a rerun hides the failed attempt.

`rust-regression` failed on attempt 1 of the push twin (run `33102414126`), in `api::dispatcher::tests::blocked_pty_worker_does_not_block_ordinary_or_different_target_progress` with `Elapsed(())` against a 1-second budget. The pull_request twin passed on the identical SHA, and attempt 2 of the push run passed on the identical SHA.

It was investigated rather than waved through, because this PR adds tests that spawn child processes to the same test binary, which is a credible contention mechanism. Measured at two exact commits, 20 full-suite iterations each with alternating order, plus 600 targeted waves forcing the suspected overlap:

- base `95db68c5`: 0/20 dispatcher failures, median 5.964ms, max 8.701ms
- head `e63a73d0`: 0/20 dispatcher failures, median 5.652ms, max 7.558ms
- forced overlap with the new startup tests: 0/600, no upward shift in the distribution

The assertion normally completes in ~5ms against 1000ms. This PR did not measurably move its failure rate or latency, and the largest local outlier occurred under `--test-threads=1`, not under parallelism.

Tracked as #1600. Not fixed here: `src-tauri/src/api/dispatcher.rs` is outside this PR's frozen scope and that fix deserves its own argument.

Issue #1700, opened 2026-09-01
IPC black box is overwritten across concurrent instances: shared localStorage and a single "main" window label


mblua:
## Summary

Two AgentsCommander builds running at the same time share one `localStorage` origin and one
window label, so the #1652 renderer black box silently overwrites itself across instances.
When one instance freezes, the other's heartbeat erases the frozen window's record, which is
exactly the evidence the instrument exists to preserve.

Found during adversarial review of #1652 phase 2. It is **not** a defect in that phase: the
frozen plan never claimed cross-instance safety, and closing it is a design change rather
than an implementation fix. Phase 2 landed with this recorded as a known limitation.

## Mechanism

Three facts combine. Each is verified in the tree:

1. **The main window's label is the literal `"main"` for every build.**
   `src-tauri/src/lib.rs`, `WebviewWindowBuilder::new(app, "main", WebviewUrl::App("index.html?window=main"))`.
   `windowType` is `"main"` too, so nothing in the record identifies which instance wrote it.

2. **The WebView2 user data folder is keyed on the Tauri bundle identifier, not on the exe
   stem.** `src-tauri/tauri.conf.json` declares `"identifier": "dev.agentscommander.app"`,
   which is identical for every suffix, so
   `%LOCALAPPDATA%\dev.agentscommander.app\EBWebView` and the `localStorage` inside it are
   shared by every build on the machine. Measured independently during #1652's phase 1
   manual gate: a testable build launched while another instance was running spawned no
   WebView2 process of its own and its renderer appeared as a child of the *first* app's
   browser process, carrying `--webview-exe-name=<the first exe>.exe`.

3. **The single-instance mutex is per-suffix** (`Local\AgentsCommander_SingleInstance_<Suffix>`),
   so two builds with different suffixes run concurrently by design rather than the second
   exiting. This is the supported side-by-side workflow, not a misuse.

## Consequences, worst first

**1. Continuous erasure of the evidence.**
Both instances write `ac.ipc.bb.cur.main` every second. If one freezes, the other's
heartbeat overwrites the frozen window's record on every tick. By the time the frozen
instance is killed there is no record of the freeze left under that key.

**2. A false previous-run verdict.**
The second instance's install rotates the first instance's *live* record to
`ac.ipc.bb.prev.main`. Its harvest then sends it, and phase 1's current-run filter does not
catch it:

```rust
if record.started_at_ms >= self.backend_started_at_ms {
    // Current run: a sibling window is still writing this one.
[...]

Issue #1652, opened 2026-08-31
UI freezes while the process stays responsive: no thread blocked in any of eight dumps, across the whole process tree


mblua:
## Summary

The window stops reacting to the mouse while the process stays alive and Windows keeps reporting it as responding. It recurs across versions: measured on **0.20.0** (2026-08-11) and again on **0.30.3.0** (2026-08-30, three times in one evening).

Five full-memory dumps have now been analysed, from two separate episodes and both processes. **No thread is blocked on a lock and no thread is spinning, anywhere, in any capture.** Every main thread sits in its normal idle wait. The freeze is not a deadlock and not a CPU-bound loop.

This was previously filed under #1298 and closed along with it. That issue fixed a real git-storm defect, but the freeze it was filed for was already shown at the time not to be caused by that storm, and it survived the fix. Reopening as its own bug.

### The dumps

| Date | Process | Version | Threads | `T0` top frame |
| --- | --- | --- | --- | --- |
| 2026-08-11 | backend (pid 45468) | 0.20.0 | 161 | `win32u!NtUserGetMessage` / `user32!GetMessageW` |
| 2026-08-11 | renderer A (pid 55052) | - | 38 | `ntdll!ZwWaitForSingleObject` |
| 2026-08-11 | renderer B (pid 45556) | - | 38 | `ntdll!ZwWaitForSingleObject` |
| 2026-08-30 | backend (pid 48396) | 0.30.3.0 | 202 | `win32u!NtUserGetMessage` / `user32!GetMessageW` |
| 2026-08-30 | backend (pid 75304) | 0.30.3.0 | 181 | `win32u!NtUserGetMessage` / `user32!GetMessageW` |

`T0` in `GetMessageW` means the Win32 message pump is alive and idle, waiting for messages it is not receiving. The Rust main thread is not blocked on anything.

Backend thread population at the 21 h 49 m dump (pid 48396, 30 PTY sessions), which is the normal idle shape:

```
131  ntdll!ZwWaitForAlertByThreadId       Rust park / SRWLock / condvar
 30  ntdll!ZwWaitForSingleObject     \ paired, one pair per PTY session
 30  ntdll!ZwReadFile                /
  6  ntdll!ZwRemoveIoCompletionEx        tokio IOCP driver
  2  win32u!NtUserGetMessage             message pump (T0, T1)
  1  ntdll!ZwWaitForWorkViaWorkerFactory
  1  ntdll!ZwRemoveIoCompletion
[...]

PR #1701, opened 2026-09-02
fix(#1652): renderer IPC black box and in-flight registry (phase 2 of 2)

mblua:
Phase 2 of 2 for #1652: the renderer half of the IPC freeze instrument. Phase 1 (backend
observer, freeze marker, black-box intake) landed in #1697 and is fully accepted, including
its manual gate.

This is not a fix. It is a diagnostic instrument that makes the next freeze identify its own
cause. This phase adds the black box in the webview's `localStorage` that phase 1's intake
harvests and classifies, which is what makes the four-way discrimination possible at all:
with no renderer record there is nothing to consume and no `VERDICT` line is possible.

Nine files under `src/`, 868 insertions, 11 deletions. No `src-tauri/` change.

## Automated gates

All run locally from the repository root.

| Gate | Result |
| --- | --- |
| `npm run typecheck` | exit 0 |
| `npm test` | 171 files, 1954 / 1954 passed, exit 0 |
| `npm run check:frontend-dependencies` | complete src root gate PASS, 360 modules, 0 errors |
| `npm run test:debt` | exit 0 |

`check:frontend-dependencies` is this phase's cycle gate and is **not in CI**, so it was run
locally and its result is reported here: zero cycles over the whole `src` root.

The thirteen tests in `src/shared/ipc-blackbox.test.ts` exist, are named as the plan writes
them, and pass. None is `.skip`, `.todo` or `.only`. A green suite alone does not satisfy
this: a file carrying nine of the thirteen would also be green, so the names were counted
and checked individually.

`npm test` emits an unrelated jsdom teardown error, `ReferenceError: document is not
defined`, from a `MutationObserver` in `src/sidebar/stores/sessions.ts`. It was confirmed
pre-existing by stashing this branch's nine files and re-running on the untouched tree. The
run still exits 0.

## Manual gates

All three acceptance criteria that require a running app were executed against a build made
from this branch's commit (`agentscommander_testeable.exe`, SHA-256
`673edcc0d9e885c5e90319ee8ac5ea8f46f967dcfeac274848dd0824ab0a381f`, 61085184 bytes, built
from `7adb1173` on a clean tree).

**AC 5, persistence: PASS.** `ac.ipc.bb.cur.main` was read directly out of the LevelDB store
and its `tickSeq` observed advancing across two independent reads (6, 59, 119). After a
force-kill of the whole process tree and a relaunch, `app.log` carried the
`[ipc-blackbox] previous-run window label=main` block. **The harvest leg is alive:**
`localStorage` survives a force-kill in this WebView2 configuration, so neither documented
negative outcome (`marker only:` plus `NO VERDICT`) occurred. The harvested block reported
`ticks=119` and `last_tick=20:44:20.194`, byte-for-byte the last record read off disk before
the kill.

**AC 6, discrimination: PASS.** Renderer-only kill, backend held alive 120.2 s, then
force-killed and relaunched. `app.log` carried the phase 1 `[ipc-observer] SILENCE` block and,
after relaunch, a `[ipc-blackbox]` block with `closed_cleanly=false` and
`VERDICT a/task-loop-stopped`.

`outlived_ms` is reported as the plan requires rather than left implicit: **2073**, against
the bound of 30000, a factor of 14.5 in margin. It is arithmetically exact, `last_tick`
20:46:54.468 minus the marker's `last_invoke` 20:46:52.395. Measured once and reported as
measured; no retry. Worth recording why the figure is trustworthy: the LevelDB store commits
on roughly a 60 s cadence, but the storage service lives in the browser process, which this
procedure deliberately leaves alive, so the last committed record landed 244 ms before the
renderer kill rather than up to a minute stale. The contrast is visible in AC 5, where the
browser process was killed too and the last record was 32 s stale. The renderer-only
procedure is the configuration in which this measurement is tight.

**AC 7, clean close: PASS.** A spec-board window was opened, closed, the app quit normally
and relaunched. `app.log` carried
`[ipc-blackbox] previous-run window label=spec-board` with `closed_cleanly=true` and
`VERDICT healthy`, and `SUMMARY 2 previous-run windows: main=healthy spec-board=healthy`.
**`pagehide` does fire on `CoreWebView2Controller` teardown in this configuration:** the
record carrying `closedCleanly:true` was written 15 ms after the `WM_CLOSE` was posted, so
the documented negative outcome did not occur.

### Disclosed deviations in the manual gates

- **`WEBVIEW2_USER_DATA_FOLDER`** was redirected to a test-owned directory. Without it the
  testable build has no WebView2 tree of its own and shares one `localStorage` origin with
  any other running instance, which is the storage these gates read. See #1700.
- **`specBoardEnabled` was flipped false to true** in the disposable testable config for
  AC 7, and restored to false afterwards. The spec-board button is gated behind
  `<Show when={settingsStore.current?.specBoardEnabled === true}>` and the setting is a
  manual opt-in with no UI (default false; no `specBoard` control exists in
  `SettingsModal.tsx`), so AC 7 is unreachable as written without it. This enables a
  documented product opt-in so the feature exists to be observed; it touches nothing the
  criterion asserts on, and `closed_cleanly` and the verdict were free to come back either
  way.
- The gates were run without mouse, without keyboard and without foregrounding, at the
  repository owner's instruction. Window interaction used the semantic `ui-*` bridge (DOM
  events, not OS input), `CloseMainWindow()` and a `PostMessage(WM_CLOSE)` to a specific
  HWND after verifying its owning PID. Foreground `HWND` was instrumented and confirmed
  unchanged across the run.

## Adversarial review

Reviewed with a battery of 24 mutants applied to `src/shared/ipc-blackbox.ts` one at a time
and reverted, against an unmutated control: 16 killed, 8 survived. Verdict PASS with no
blocking findings. Eight non-blocking findings were recorded; the significant one is filed
as #1700 and is a pre-existing design gap rather than a defect in this phase.

## Known errata in the frozen plan, carried unchanged

None changes a decision, constant, threshold, field, verdict or test. They are recorded
rather than fixed, because the plan is frozen and editing it would invalidate its approval.

- `epic.md:37` says the funnel carries 149 of 151 call sites. The funnel object carries 148,
  the file carries 149, and the sentence is self-inconsistent because it also names three
  leaks.
- `epic.md:106`'s discrimination-table row omits the `OVERDUE_MS` half of the predicate. The
  row is wider than the code; the code is on the safe side.
- `epic.md:426` and `phase-2:680` carry stale motivation clauses; the second cites the epic
  for a position the epic later abandoned.
- `phase-2:158`'s direction word, and two bare "23rd field" survivors.

Two further errata were found during implementation, both in acceptance criterion 3, and
both because the plan's own mandated text makes the literal unreachable:

- **The surviving line is `src/shared/transport-tauri.ts:46`, not `:45`.** The plan mandates
  a value import from `./ipc-blackbox`, and the only import above line 45 at the base SHA is
  `import type`, which cannot absorb a value import. Every conforming placement adds one line
  above 45, so `:45` is unreachable short of deleting an unrelated line.
- **The unfiltered grep prints 6, not 5.** The plan's own verbatim replacement comment
  carries the literal string `` `@tauri-apps/api/core` `` and the grep matches it. 7 at the
  base, minus the two removed production sites, plus that mandated comment, is 6.

The load-bearing half of AC 3 held exactly: the filtered grep prints `1`, down from `3` at
the base SHA, and the sole surviving production site is the Tauri transport's own dynamic
import. AC 4 prints nothing and exits 1.

## Follow-up filed

#1700: the black box is overwritten across concurrent instances, because two builds share
one `localStorage` origin (WebView2 keys its user data folder on the bundle identifier, not
the exe stem) and one `"main"` window label. Single-instance operation is unaffected. Closing
it is a design change, not an implementation fix, and no gate in either phase can catch it
because all of them run one instance.

Refs #1652

https://claude.ai/code/session_01VB6gNKEYM4TLCnUwBJi1Ts


mblua:
## SonarCloud Quality Gate: failed, merging anyway, with the reason recorded

`SonarCloud Code Analysis` reports **Quality Gate failed: B Security Rating on New Code
(required >= A)**. It passed on phase 1 (#1697), so this is a genuine regression of that
gate introduced here, not a flaky check. Recording the full analysis rather than merging
past it quietly.

### The finding

One issue, and it is **not** a false positive.

```
[MINOR] VULNERABILITY  tssecurity:S8475
src/shared/ipc-blackbox.ts:215
Ensure that tainted data is sanitized before being written to browser storage.
```

The traced flow is real. `windowType` is taken from the query string at
`src/shared/ipc-blackbox.ts:250`:

```js
windowType = new URLSearchParams(window.location.search).get("window") ?? "main";
```

and reaches storage unvalidated at `:215`:

```js
localStorage.setItem(storageKey, JSON.stringify(buildRecord(Date.now())));
```

So `location.search` flows into `localStorage`. Sonar is right about the flow.

### Why it is being merged rather than fixed here

**Practical risk is low.** In the Tauri app the window URL is constructed by the Rust
backend (`index.html?window=main`), so it is not attacker-controllable. In the browser
client the URL *is* controllable, but `src/main.tsx` returns `BrowserApp` before the harvest
line, so a record written there is never harvested and never reaches `app.log`. The
realistic worst case is forged `[ipc-blackbox]` lines in `app.log`, and producing it
requires local code execution the attacker would already need.

**The flagged line is mandated verbatim by the frozen plan**, at
`phase-2-frontend-ipc-blackbox.md:237` step 4. Fixing it at the source would edit text the
plan mandates, which invalidates the plan's approval and all three reviewer verdicts and
sends the task back to a new human gate. That is the same bind as the two AC 3 errata above.

**It is not a required check.** All 13 required status checks on `main`
(`validate-branch-name`, `lockfile-drift`, `rust-regression`, `frontend-regression`,
`rust-regression-linux`, `rust-regression-macos`, the four `terminal-snapshot-portable`
legs, `windows-release-cli-smoke`, `test-debt`, `rust-fmt`) are green.

### Where it gets fixed

Folded into #1707 as an explicit requirement. That issue already reworks the install path
where line 250 lives, so sanitizing `windowType` there is a change to code that is being
rewritten anyway, at the natural moment, without deviating from any frozen plan.

**Known consequence, stated so it is not a surprise:** `main` will carry a B security rating
on new code until #1707 lands.

Decision made by the repository owner with the above on the table.

Issue #1715, opened 2026-09-02
config.json is rewritten and fsynced on every armed idle edge, including no-op writes


mblua:
Found in the code review for #1682, accepted there as a real cost rather than a blocker.

## What happens

`update_config_json_object` (`src-tauri/src/config/local_config_io.rs:11-77`) serializes, `File::create` + `write_all` + `flush` + `sync_all`, then publishes via a temp file + `ReplaceFile` with up to 4 attempts and 260 ms of backoff. It does this **unconditionally**, including when the mutate closure changed nothing.

`set_last_agent_message_at` (`src-tauri/src/config/agent_config.rs:179-207`) has a monotonic early `return Ok(())` for a value that is not newer than the stored one. That early return still costs a full durable rewrite of `config.json`, because the decision is made inside the closure and the write happens regardless.

## Why it is worth looking at now

Before #1682 the writers of this file ran roughly once per launch (`set_last_coding_agent`). #1682 writes on **every busy-to-idle edge of every armed agent session**, which is orders of magnitude more often.

Two consequences:

1. **Same-process is safe and I checked it.** `local_config_write_lock` is a process-global `Mutex` and the read-modify-write happens under it, so two sessions settling at once for the same replica cannot tear the file or drop a neighbouring `tooling` key.
2. **Across two AC instances sharing a replica the read-modify-write is not atomic**, and it is last-writer-wins on the whole file. That race predates #1682; #1682 raises the collision rate substantially.

The write is also synchronous, on a tokio worker, inside the spawned idle task while the `SessionManager` read guard is held (`lib.rs:2494-2510`).

## Suggested direction

Give `update_config_json_object` a way for the mutate closure to report "nothing changed" and skip the write entirely. That is a change to shared config-write code touching several callers, which is why it was not folded into #1682's landing.

The cross-instance atomicity question is separate and larger.

Issue #1714, opened 2026-09-02
TerminalView: onMount's async chain can outlive the component, leaking listeners


mblua:
Found while landing #1682, and deliberately left out of that delivery's scope.

## The defect

`TerminalView`'s `onMount` body is `async` and registers several listeners in sequence, each behind an `await`. The component can be disposed while that chain is suspended: `shouldMountTerminal` drops `TerminalView` whenever the selection leaves live mode, which is frequent and needs no window close.

When that happens, `onCleanup` has already run past every unlisten holder below the suspension point. Any registration that resolves afterwards attaches to a dead view and is never released.

## What #1682 already fixed, and what it did not

#1682 added one more `await` to that chain and, in `f8d17b6a`, closed the race for its own registration: an `unmounted` flag set as the first statement of `onCleanup`, checked immediately after the `onSessionAgentMessage` registration, which releases the unlisten and returns early from `onMount`.

That early return also protects everything below it (`onTerminalDetached`, `onCloseRequested`). **The registrations ABOVE it still carry the race**: `onPtyOutput` and `onSessionDestroyed`.

## How it shows up

It is observable in the vitest suite. Because `renderWithFakeTransport(...).cleanup()` calls `dispose()` and `restoreTransport()` synchronously, a registration resolving after teardown talks to the real Tauri transport and rejects with `TypeError: Cannot read properties of undefined (reading 'transformCallback')`. That surfaces as an unhandled rejection: every test passes and `npm test` still exits 1, attributed to whichever file vitest happens to blame.

Reproduction before the #1682 guard was a single file, ~8s:

```
npx vitest run src/terminal/components/TerminalView.attachment.test.tsx
```

Note the `ReferenceError: document is not defined` from `sidebar/stores/sessions.ts:52` prints in both green and red runs and is unrelated.

## Suggested fix

Apply the same guard to the remaining registrations, or hoist the pattern. `terminal/App.tsx:69-75` already has the house idiom for this: a `disposed` flag with an `addUnlistener` helper that calls the unlisten immediately when the component is gone.

## Why it was not fixed in #1682

It predates #1682, and widening the last phase of a five-phase delivery to cover a pre-existing race was the wrong trade at that moment.

Issue #871, opened 2026-07-08
bug: Restart Session fresh intent cleared by non-substantive PTY writes → coordinator resumes old session after app restart (#756 residual 1)


mblua:
## Summary

Applying **"Restart Session"** to a coordinator, then closing and reopening the whole app, resurrects the pre-restart conversation (Claude `--continue` is injected). This is the documented **residual 1** of #756 (fixed via PR #764 / `f5e8635`), **not** a regression of #756's durable-intent mechanics — those still work. A separate, too-aggressive re-arm path defeats them in practice.

## Repro (verified on a build from 2026-07-08, which already contains the #756 fix)

1. Apply "Restart Session" to a coordinator (e.g. tech-lead). The fresh session comes up clean (empty prompt, context 0%).
2. Close the whole app, reopen, return to the coordinator → **the pre-restart session resumes** (Claude Code shows *"This session is Xh old … Resume from summary / Resume full session"*).

## Root cause (dev-rust investigation, with live log evidence from the affected machine)

Restart Session correctly stamps the durable fresh intent in **both** places:
- session record `start_fresh_on_restore = true` (`sessions.json`)
- coordinator-clock mirror `start_fresh_at` (`coordinator_clocks.json`)

But within **~3 seconds** of the restart, a user-message path re-arms the intent back to `false`, **before the app is even closed**:

- `app.log` 11:29:56 — restart: `skip_auto_resume=true -> inject_continue=false` (fresh, correct)
- `app.log` 11:29:57 — mirror stamped: `start_fresh_at set for '…/tech-lead' (#756)`
- `app.log` 11:29:59 — **`start_fresh_on_restore true -> false (user message, re-armed)`** ← intent dies here
- `app.log` 13:37:20 — later restore: `skip_auto_resume=false -> inject_continue=true` → `Auto-injected --continue`
- `app.log` 13:37:56 → 13:37:59 — repro repeated: restart fresh, then `user message, re-armed` 3s later.
- Current `sessions.json` row: `startFreshOnRestore: false`; current `coordinator_clocks.json` row: no `startFreshAt` key.

The re-arm fires via `pty_write` → `note_user_message_to_session` on **any** bytes written to the PTY, including **non-substantive terminal writes** (control/focus sequences, terminal init, empty Enter, navigation keys) — not only real user prompts. So a freshly restarted, untouched coordinator loses its fresh intent within seconds, defeating the #756 fix.

## Code chain (current `origin/main` @ `4a2810f`)

[...]

PR #1716, opened 2026-09-02
feat(#1682): show the last coding-agent message timestamp in the terminal status strip

### LINKED ISSUE #1682: feat: show the last coding-agent message timestamp in the terminal status strip
## Situation

The terminal view renders a small status chip at the bottom-right of the terminal container
showing only the grid size:

- `src/terminal/components/TerminalView.tsx:259` sets that chip text from a COLS/ROWS template literal
- class `terminal-grid-status`, `data-testid="terminal-grid-size-<sessionId>"`,
  absolutely positioned `right: 8px; bottom: 4px`.

Nothing in that strip, or anywhere else in the terminal view, tells the user *when* the coding
agent last spoke on that terminal. Coming back to a replica after a while, or reopening the app,
there is no way to tell whether the output on screen arrived a minute ago or two days ago.

The only existing "activity" stamp is `lastActivityBySessionId` in
`src/sidebar/stores/sessions.ts:678`, which records `performance.now()` (a monotonic clock, not a
wall clock) purely to order rows in the sidebar. It is in-memory, resets on every app start, and
cannot serve this purpose.

## Why the change is needed

Requested by the user: the terminal status strip must show the system date and time of the last
message received from the coding agent on that terminal, immediately to the left of the existing
`COLS/ROWS` chip. The value must be **persisted**, because an in-memory stamp would be blank
exactly when it is most useful, right after reopening the app.

## How it will be changed

**Signal - end of response, not streaming.** The stamp advances when the coding agent finishes
responding, i.e. on the existing busy -> idle transition, not on every chunk of PTY output. The
backend already emits that event from the idle detector (`src-tauri/src/lib.rs:2457-2470`, event
`session_idle`), and the frontend already consumes it (`src/sidebar/App.tsx:773`). Using it avoids
having to intercept every PTY write and avoids a high-frequency persistence path. History replay
on re-attach, AC-generated banners and reconnect notices must not advance the stamp.

**Persistence - per agent/replica, in the `tooling` block.** A new field alongside the existing
`lastCodingAgent` / `codingAgents[<id>]` entries in `config.json`, written through
`src-tauri/src/config/agent_config.rs` (see `set_last_coding_agent`, `agent_config.rs:132`, called
from `src-tauri/src/commands/session.rs:2509`). Scope is the agent/replica path, so the value
survives both closing the app and restarting the session. Timestamps are stored RFC3339/UTC to
match the existing `lastUsed` convention.

`codingAgents[<id>].lastUsed` must **not** be overloaded for this: it means "last time this coding
agent was launched", not "last output received". A distinct field is required.

**Display.** A new element in the same bottom strip, to the left of the grid-size chip, rendering
the stored timestamp in **local time** with the mask `MM-DD HH:MM`, for example:

```
08-31 21:29   COLS: 163 ROWS: 44
```

When no timestamp has been recorded yet for that agent/replica, the new element renders nothing;
only the `COLS/ROWS` chip is shown.

**Do not break existing readers of the `tooling` block:**

- `src-tauri/src/commands/ac_discovery.rs` (reads `tooling.lastCodingAgent`)
- `src-tauri/src/cli/list_peers.rs` (exposes/excludes `lastCodingAgent` in CLI JSON; has tests
  asserting the `"lastCodingAgent":` key)
- `src-tauri/src/cli/send.rs` (`auto` mode picks the saved `lastCodingAgent`)
- `src-tauri/src/cli/self_restart.rs`

Note that every AC session restart already rewrites the `tooling` block
(`execute_restart_transaction` -> `create_session_inner_impl` with `skip_tooling_save = false`,
`src-tauri/src/commands/session.rs:4116`), so no acceptance criterion may assert that `config.json`
is byte-identical across a restart.

## Acceptance criteria

1. After a coding agent finishes a response on a terminal, the strip shows a `MM-DD HH:MM` local
   timestamp to the left of the `COLS/ROWS` chip.
2. The value survives closing and reopening the app, and survives a session restart, for the same
   agent/replica.
3. Streaming output alone does not advance the stamp; only the completion of a response does.
4. Re-attaching a terminal (history replay) does not advance the stamp.
5. With no recorded value, only the `COLS/ROWS` chip renders.
6. Existing `tooling`-block readers and their tests keep passing.

## Open for the architecture plan

- Exact JSON key name and its placement within `tooling`.
- The write path from the idle event to `config.json`, and any write-coalescing policy.
- Whether the element is a sibling of the grid chip or a second span inside a shared container,
  and the corresponding `data-testid` / `aria-label`.
- Test coverage, including the existing `src/terminal/components/TerminalView.spawn-size.test.tsx`
  and `src/shared/terminal-viewport.test.ts`.

mblua:
Adds the timestamp of the last message received from the coding agent to the terminal status strip, immediately left of the existing `COLS/ROWS` chip, masked `MM-DD HH:MM` in local time, persisted per agent/replica.

```
08-30 22:29   COLS: 95 ROWS: 34
```

With no recorded value the element renders nothing: verified by pixel-diffing two captures over the strip band, where every column from the chip's left edge rightwards is byte-for-byte identical with and without the stamp.

## How it works

- **Signal:** the busy-to-idle edge of an *armed* session, not per-chunk streaming. A turn arms when a message write reaches one of three arming sites; three clear sites pair with them.
- **Storage:** `tooling.lastAgentMessageAt` in the replica's `config.json`, so it survives app close and session restart.
- **Suppression:** three gates. An armed session with an `agent_id` only; no recent unsubmitted user typing; and no recent control write of ours re-opening the busy period.
- **Frontend:** a right-anchored strip containing both elements, with a per-entry monotonic watermark so an older event or a late hydration cannot move the value backwards.

## Known and disclosed

The stamp keys on **arming**, not on proven delivery. Three residuals ship deliberately:

- **R1.** An edge landing outside the 6000 ms user-write window can move the stamp on something that is not a message. The reachable cases are a window resize or a session switch whose repaint outlives `resize_grace`, and a slow-repainting child after a focus change. (An earlier draft of this note also named auto-title arming; that path is dead code twice over and does not arm.)
- **R7.** `session/manager.rs` arms beside `notify_pty_input_busy`, before the write is proven submitted.
- **R8.** `pty/inject.rs` arms on a path whose `send_enter` is decided from a stem-only shell profile.
- A line typed and then backspaced away leaves `pending_nonspace` set (pre-existing #871 imprecision), so it can suppress for up to 6000 ms after the last contributing keystroke. Bounded, not unbounded.
- An editing burst begun while the agent is still producing output, with no 2500 ms gap inside it and more than 6000 ms of non-contributing keys, can let the stamp land on the user's own editing echo.

Two follow-ups filed rather than folded in: #1714 (a pre-existing disposal race on the `onMount` registrations above this change's guard) and #1715 (`config.json` rewritten and fsynced on every armed idle edge, including no-op writes).

## Verification

- `cargo fmt --check`, `cargo check --all-targets`, `cargo clippy --workspace --all-targets -- -D warnings`: exit 0.
- `cargo test --lib --bins --tests --no-fail-fast`: exit 0, 3951 lib tests green.
- `npm test`: 173 files, 1972 tests, exit 0, zero unhandled errors. `npm run typecheck`: exit 0.
- `npm run check:frontend-dependencies` (a real gate that no CI job runs): complete-root `modules: 364, errors: 0`, PASS.
- Module cycle detector: 1 cycle, SCC `40d22fc71179d71f`, size 86, identical to `origin/main` measured alone. `module-arcs.txt` regenerated, 1080 arcs.
- Adversarial review found one blocker, an aging gate that did not age out: any terminal traffic refreshed its clock, so an abandoned half-typed line could suppress the stamp indefinitely. Fixed in `c84496e2` by aging from contributing chunks only. The fix was verified by a differential run of the old and new classifiers over 2,066,074 chunks from both starting states, 4,132,148 comparisons, zero divergence in `submitted` or `pending_nonspace`.
- Visual gate on a built binary: a seeded value renders correctly across a date-and-hour timezone boundary, an absent value renders nothing without moving the chip, a driven turn moves it live and persists, and both values rehydrate after a full app restart.

Closes #1682

https://claude.ai/code/session_017rWyUN1fJ7Vi1UTDySz524


mblua:
## Merging with admin over a red `rust-regression`, with the evidence

`rust-regression` is red on this head. It is not this change. The control that settles it:

**`feature/1691-frontend-cancellation-and-outcomes` at `d52f8738` — unrelated work, 3952 tests, none of them ours — failed `agent_version::tests::probe_version_skips_an_ansi_only_first_line` at `agent_version.rs:1417` at 13:51Z, the same window.** `main` itself failed the same test at the same line at 05:05Z (commit `48783767`, job 100131551664).

Five runs of this commit produced four different failing sets:

| run | failed |
|---|---|
| c84496e2 push | `agent_version` (`timed out after 15s (killed)` spawning powershell.exe) |
| c84496e2 PR | none — passed the tests the twin failed, same SHA |
| 5f501e4c (a) | `agent_version` |
| 5f501e4c (b) | `agent_version` + 2x `session::selection::tests::capped_shutdown_abort_joins_*` |
| rerun (a) | `pty::spawn_diagnostics::tests::monitor_writes_the_startup_verdict_for_a_silent_child_and_keeps_watching` |
| rerun (b) | `agent_version` |

Different sets across runs of identical bytes is environmental, not a regression: a real regression fails the same test every time. All of them are wall-clock tests against real process spawns.

**Locally, on this exact merged tree:** those three tests pass under `--test-threads=1`, and the full suite is 4243 passed / 0 failed across 46 binaries. `cargo fmt --check`, `cargo clippy -D warnings`, `npm test` (173 files, 1972 tests), `npm run typecheck` and `npm run check:frontend-dependencies` (modules 364, errors 0) are all green.

**This change touches none of the implicated code.** `agent_version.rs` and `pty/spawn_diagnostics.rs` are untouched. The entire diff to `session/selection.rs` is one line: `agent_turn_armed: false,` in a struct literal.

Job durations rule out a load hypothesis: these runs took 13-16 min against 16-18 min for green runs on `main`.

Filed separately for the flaky test.

Issue #1730, opened 2026-09-02
Sidebar rows: Telegram glyph instead of the bridge dot, agent name as a chip, and a new chip order


mblua:
Three cosmetic changes to the sidebar rows, applied to **both** row kinds: the replica / orchestrator row (`ProjectPanel.tsx` `renderReplicaItem`) and the plain session row (`SessionItem.tsx`). No behaviour change: the Telegram attach/detach logic, the bridge store and the badge data sources all stay exactly as they are. This is presentation only.

No chip is added to either row beyond `agent-name-chip`, and no chip is removed. Each row keeps exactly the chips it renders today; the target order is applied to each row over the subset it actually has.

**Before** (orchestrator row of `room-12-ac-dev-team-v4`, current `main` @ `ac845616a9c8b4f82684499a5727a70aeaef39d1`):

```
        Clean                                            <- task title line
        ac-tech-lead-v4@AgentsCommander_iac              <- name line, deleted by part 2
(dot)   [0m] [AgentsCommander/main] [personal/main]
        [agentscommander_webpage/main] [Claude Code] [B]
        [CTX N/A] [ORCHESTRATOR] [ROOM-12-AC-DEV-TEAM-V4]        (bridge dot)
```

**After** (same row, same state):

```
        Clean                                            <- task title line, unchanged
(dot)   [0m] [ac-tech-lead-v4] [ORCHESTRATOR] [Claude Code]
        [B] [CTX N/A] [ROOM-12-AC-DEV-TEAM-V4]
        [AgentsCommander/main] [personal/main]
        [agentscommander_webpage/main]                    (Telegram glyph)
```

---

## 1. Replace the Telegram bridge dot with the Telegram glyph

Today an attached Telegram bridge is signalled by a 6px coloured circle, `.session-item-bridge-dot`, tinted with the bot colour. A bare dot carries no meaning on its own: it is indistinguishable from the row status dot on the left of the same row.

Show the Telegram symbol instead, still tinted with the bot colour (`bridge().color`) and still carrying the same `title={`Telegram: ${bridge().botLabel}`}`.

`TelegramIcon` already exists at `src/sidebar/components/TelegramIcon.tsx` and is the icon used by the attach/detach button; reuse it, do not add a second SVG.

Call sites of the dot:

- `src/sidebar/components/ProjectPanel.tsx:2545` — replica / orchestrator row. This row has **no** Telegram button, so the dot is the only Telegram affordance. Here the dot must become the icon.
[...]

Issue #1723, opened 2026-09-02
Show the Ctrl+Shift+W shortcut on the close-session controls


mblua:
## Situation

AC binds `Ctrl+Shift+W` to "close the currently selected session" in `src/shared/shortcuts.ts:13-25`, which calls `requestCoordinatorCloseById`. Three UI controls run the very same close path:

- The replica context-menu item labelled `Close Session` (`src/sidebar/components/ProjectPanel.tsx:3660` -> `handleReplicaContextClose` at `:1486` -> `requestCoordinatorClose`).
- The `X` close button on a session row (`src/sidebar/components/SessionItem.tsx:492`, tooltip `title="Close session"`).
- The `X` close button on the root-agent banner (`src/sidebar/components/RootAgentBanner.tsx:557`, tooltip `title="Close session"`).

None of the three mentions the keyboard shortcut. Today the only place that documents it is `docs/reference/keyboard-shortcuts.md:9`, so a user working in the app has no way to discover that the same action has a shortcut.

## Why the change is needed

The shortcut already exists and is already the fastest way to close a session, but it is invisible from the UI. Surfacing it next to the controls that perform the identical action is the standard discoverability affordance and costs nothing at runtime.

## How it will be changed

Copy-only change, no behavior change, no backend work:

1. `src/sidebar/components/ProjectPanel.tsx:3660` - the menu item text becomes `Close Session (Ctrl+Shift+W)`.
2. `src/sidebar/components/SessionItem.tsx:493` - the tooltip becomes `Close session (Ctrl+Shift+W)`.
3. `src/sidebar/components/RootAgentBanner.tsx:558` - the tooltip becomes `Close session (Ctrl+Shift+W)`.
4. `src/sidebar/components/ProjectPanel.context-menu.test.tsx` - the exact-text menu lookups at lines 553, 1441, 2048 and 2069 must be updated to the new label.
5. `docs/reference/keyboard-shortcuts.md` - note that the close-session controls surface the shortcut.

## Known nuance (documented, not fixed here)

`Ctrl+Shift+W` closes the **selected** session, while the context-menu item closes the session the menu was opened on. When the user right-clicks a session that is not the selected one, the two targets differ. This is the same convention most desktop apps use and is out of scope for this issue.

[...]

Issue #1719, opened 2026-09-02
ui(sidebar): aesthetic details backlog for the replica context menu


mblua:
Running list of small aesthetic corrections to the replica context menu in the sidebar. Items are appended as they come up; each one is self-contained and can land on its own branch. Origin of this surface: the icon and colour pass in #1708.

---

### Current state

| Row | Icon today | Where |
| --- | --- | --- |
| `Add to Group` | emoji 👥 `&#x1F465;` in the 16px gutter | `src/sidebar/components/ProjectPanel.tsx:1710` |
| `Create new group` | **no icon at all**, the gutter slot is missing entirely | `src/sidebar/components/ProjectPanel.tsx:1642-1652` |

Both rows live in `renderAddToGroupItem`, which is rendered from the active-replica menu (`ProjectPanel.tsx:3722`) and the inactive-replica menu (`ProjectPanel.tsx:3838`), so there is a single source for each.

### Wanted

1. `Create new group` takes the 👥 emoji `&#x1F465;` that `Add to Group` carries today.
2. `Add to Group` takes a `user-plus` SVG, tinted with the `--status-waiting` token (green).

Rationale: `user-plus` reads as "add someone to this", which is what `Add to Group` does. 👥 reads as "a group", which is what `Create new group` produces.

### The SVG

This is the exact shape that was reviewed and approved. Lucide `user-plus`, stroke 2, `viewBox="0 0 24 24"`, rendered at 14px.

```svg
<svg viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2"
     stroke-linecap="round" stroke-linejoin="round" aria-hidden="true">
  <path d="M16 21v-2a4 4 0 0 0-4-4H6a4 4 0 0 0-4 4v2"/>
  <circle cx="9" cy="7" r="4"/>
  <line x1="19" y1="8" x2="19" y2="14"/>
  <line x1="22" y1="11" x2="16" y2="11"/>
</svg>
```

Note a deliberate deviation from house style: `DetachIcon` and `ReattachIcon` are **solid** Heroicons (`fill="currentColor"`), while this one is a **stroked** icon. That is the version that was visually approved at the menu's real size; do not silently substitute a solid variant.

Ship it as its own component next to the existing ones, e.g. `src/sidebar/components/UserPlusIcon.tsx`, following the `DetachIcon.tsx` pattern: a `Component<{ class?: string }>` that forwards `class` to the `<svg>` so the caller controls size and tint.

[...]

Issue #1708, opened 2026-09-02
Sidebar context menu: give Add to Group an icon, colour the Telegram / detach / Edit TASK title entries, and rename Open in new window to Detach session


mblua:
## Situation

The session context menu in the sidebar mixes coloured and uncoloured icons without a rule, and one entry has no icon at all. All anchors below were verified against `main` at `b7453a0`.

1. **`Add to Group` has no icon.** The icon span is rendered but left empty, so the label sits misaligned against every other entry in the menu:

   ```tsx
   // src/sidebar/components/ProjectPanel.tsx:1708
   <span class="session-context-option-icon" aria-hidden="true" />
   ```

2. **The Telegram icon is grey until a bridge exists.** It only takes a colour when a bridge is active, from the bot's own configured colour. With no bridge the fallback is an empty style object, so `Attach Telegram` inherits the plain menu foreground and reads as disabled:

   ```tsx
   // src/sidebar/components/ProjectPanel.tsx:3682
   style={telegramBridge() ? { color: telegramBridge()!.color } : {}}
   ```

3. **`Open in new window` is misnamed and its icon does not read.** The action moves the session out of the main window; it does not open a second view of it. The current pair is also asymmetric: `Open in new window` going out, `Re-attach to main` coming back. The icon is `⧉` (U+29C9), a typographic glyph rather than an interface icon, which at the menu's 13px reads as a smudge.

4. **`Edit TASK title` is uncoloured** while sitting directly above `Clear task title`, which is a fixed-colour emoji. The two operate on the same piece of state but do not read as a pair.

## Why this needs to change

The menu already carries colour meaning, it just carries it by accident: `#ef4444` marks session lifecycle (`Restart Session`, `Close Session`, `Open Matrix folder`) and `#a855f7` marks the filesystem (the repo entries). Everything else is grey. Making the remaining entries follow the same rule turns colour into a category the user can scan, and gives future menu entries a lane to land in instead of another ad-hoc decision.

`Add to Group` is a plain bug: the empty span breaks the icon column alignment.

### 1. `Add to Group` gets an icon

`ProjectPanel.tsx:1708` renders `&#x1F465;` (👥, U+1F465) inside the existing span.

### 2. Telegram falls back to the Telegram blue

[...]

PR #1740, opened 2026-09-03
ui(sidebar): swap the Add to Group and Create new group icons (refs #1731)

### LINKED ISSUE #1731: ui(sidebar): swap the Add to Group and Create new group icons
Child of #1719, Item 1. Self-contained; lands on its own branch.

## Situation

In the sidebar replica context menu, `renderAddToGroupItem` (`src/sidebar/components/ProjectPanel.tsx:1689`) renders two rows:

| Row | Icon today | Where |
| --- | --- | --- |
| `Add to Group` | emoji 👥 `&#x1F465;` in the 16px gutter | `ProjectPanel.tsx:1710` |
| `Create new group` | **no icon at all**, the gutter slot is missing entirely | `ProjectPanel.tsx:1642-1652` |

The helper is rendered from the active-replica menu (`ProjectPanel.tsx:3722`) and the inactive-replica menu (`ProjectPanel.tsx:3838`), so there is a single source for each row.

## Why the change is needed

The glyphs are swapped relative to what they mean. 👥 reads as "a group", which is what `Create new group` produces, not what `Add to Group` does. `Add to Group` needs an icon that reads as "add someone to this". Separately, `Create new group` has no gutter at all, so its label does not line up with the icon rows around it.

## How it will be changed

1. `Create new group` takes the 👥 emoji `&#x1F465;` that `Add to Group` carries today, inside a `.session-context-option-icon` gutter.
2. `Add to Group` takes a `user-plus` SVG, tinted through a class whose rule is `color: var(--status-waiting);`.

### The SVG

Exact shape reviewed and approved. Lucide `user-plus`, stroke 2, `viewBox="0 0 24 24"`, rendered at 14px.

```svg
<svg viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2"
     stroke-linecap="round" stroke-linejoin="round" aria-hidden="true">
  <path d="M16 21v-2a4 4 0 0 0-4-4H6a4 4 0 0 0-4 4v2"/>
  <circle cx="9" cy="7" r="4"/>
  <line x1="19" y1="8" x2="19" y2="14"/>
  <line x1="22" y1="11" x2="16" y2="11"/>
</svg>
```

Deliberate deviation from house style: `DetachIcon` and `ReattachIcon` are **solid** Heroicons (`fill="currentColor"`); this one is **stroked**. That is the version visually approved at the menu's real size; do not silently substitute a solid variant.

Ship it as its own component next to the existing ones, `src/sidebar/components/UserPlusIcon.tsx`, following the `DetachIcon.tsx` pattern: a `Component<{ class?: string }>` that forwards `class` to the `<svg>` so the caller controls size and tint.

## Constraints that must not be violated

1. **Colour comes from a token, never a hex literal.** `#22c55e` is only the dark value of `--status-waiting`. This is the rule #1708 established and `src/sidebar/styles/context-menu-icon-css.test.ts` exists to enforce it.
2. **The gutter has no spare room.** `.session-context-option-icon` is 16px wide with a 13px font (`src/sidebar/styles/sidebar.css:3350`), and its comment states no glyph may grow the box or paint it without widening it past the widest glyph first. Size the new SVG at 14px, exactly like `.session-context-detach-icon` does.
3. **An emoji ignores `color`.** 👥 is painted by the system emoji font. Do **not** add a token colour class to the `Create new group` gutter: it would be dead CSS that reads as an intent the page never honours.

## Gutter width decision

The group rows inside that flyout use `.session-context-option-check` (14px), not `.session-context-option-icon` (16px). `Create new group` uses the standard **16px** `.session-context-option-icon`, consistent with every other icon row in the menu, accepting that its label then sits 2px to the right of the group names above it. If that offset reads badly in the built app, the follow-up is to move the create row to the 14px width, not to change the shared 16px slot.

## Disclosed contrast shortfall

Computed (not measured in-app) WCAG contrast for the `user-plus` icon against the context-menu surface:

| Theme | Icon | Surface | Ratio |
| --- | --- | --- | --- |
| Dark | `--status-waiting` `#22c55e` | `--sidebar-surface` `#0e0e18` | **8.42 : 1** |
| Light | `--status-waiting` `#16a34a` | `--sidebar-surface` `#ebebf0` | **2.77 : 1** |

The light value falls short of the 3:1 non-text minimum. This is the same class of known, disclosed shortfall that #1708 accepted for the light amber (2.47 on the same surface), and it is recorded here rather than silently shipped. Retuning the light `--status-waiting` is a legitimate follow-up; it must not be done by hard-coding a darker green at the call site, which would defeat constraint 1.

## Acceptance criteria

1. `Add to Group` renders the `user-plus` SVG inside `.session-context-option-icon`, with no text glyph left in that gutter.
2. The SVG is tinted through a class whose rule is `color: var(--status-waiting);` and which contains **no** hex literal, mirroring `.session-context-detach-icon` and `.session-context-task-icon`.
3. `Create new group` renders `&#x1F465;` inside a `.session-context-option-icon` gutter, and carries no colour class.
4. `src/sidebar/components/ProjectPanel.context-menu.test.tsx:1455` is updated: it currently asserts `Add to Group`'s gutter `textContent` is `"\u{1F465}"`, which this change makes false. It must instead assert that `Add to Group` contains the SVG (by its tint class) and that the `Create new group` button, reachable via its `replica.<wg>.groups.create` testid, carries `"\u{1F465}"`.
5. `src/sidebar/styles/context-menu-icon-css.test.ts` gains a fourth case pinning the new tint class to `color: var(--status-waiting);` with no hex literal, and `--status-waiting` is added to the token list in its light-override case (it already has a distinct light value, `#16a34a` vs `#22c55e`).
6. Both menus that render `renderAddToGroupItem` (active and inactive replica) show the change, since they share the helper.
7. Verified visually in the built app in **both** themes, not only in jsdom, which never applies these stylesheets.

Parent backlog: #1719

mblua:
Item 1 of the context-menu aesthetic backlog #1719: the `Add to Group` and `Create new group` icons were swapped relative to what they mean, and `Create new group` had no gutter at all.

- `Add to Group` now renders a stroked Lucide `user-plus` SVG, tinted through `.session-context-group-add-icon` whose rule is `color: var(--status-waiting);` with no hex literal.
- `Create new group` takes the 👥 `&#x1F465;` emoji that `Add to Group` used to carry, inside a standard `.session-context-option-icon` gutter, with no colour class (an emoji ignores `color`).

Both rows come from one helper, `renderAddToGroupItem`, so the active-replica and inactive-replica menus change together.

## Files

| File | Action |
| --- | --- |
| `src/sidebar/components/UserPlusIcon.tsx` | added |
| `src/sidebar/components/ProjectPanel.tsx` | modified |
| `src/sidebar/styles/sidebar.css` | modified |
| `src/sidebar/components/ProjectPanel.context-menu.test.tsx` | modified |
| `src/sidebar/styles/context-menu-icon-css.test.ts` | modified |

Zero Rust files, zero config, workflow, lockfile or dependency changes.

## Verification, measured on the merged tip `baf5ee10`

| Gate | Result |
| --- | --- |
| `npm run typecheck` | exit 0, no diagnostics |
| `npx vitest run` (the four-file selection) | `Test Files 4 passed (4)`, `Tests 97 passed (97)`, exit 0 |
| `npm test` | exit 0, `174 passed (174)` files, `2041 passed (2041)` tests |
| `npm run check:frontend-dependencies` | `modules: 366, errors: 0, dependencies: 1615`, `complete-root gate: PASS` |
| `git status --porcelain` | empty |

`check:frontend-dependencies` is wired into no workflow, so no CI job re-derives it. It was measured twice independently, by the implementer and by the reviewer in a separate tree: `origin/main` alone is `365 / 0 / 1613`, so this change contributes exactly +1 module and +2 arcs. `UserPlusIcon.tsx` has in-degree 1 and out-degree 1 into an external type declaration with no path back into `src/`, so it is a singleton SCC and cannot lie on a cycle.

## Review

Adversarial review passed on the implementation commit. The reviewer ran eight mutations against the four assertions this change adds or modifies, and all eight were killed:

- reverting either production half fails exactly one test, the intended one;
- leaving a stray `&#x1F465;` beside the SVG fails the `textContent === ""` assertion;
- adding a colour class to the `Create new group` gutter fails its `toBeNull()`;
- swapping the token to `--status-pending` with no hex fails the CSS `toContain`, which the hex regex alone cannot see;
- deleting the light `--status-waiting` override fails the light-override case, so the added token is not decorative.

Neither modified assertion is weaker than what it replaced: one went from a single assertion to two strictly stronger ones, the other from two tokens to three.

An uncaught jsdom `ReferenceError: document is not defined` from `src/sidebar/stores/sessions.ts:52` appears in the full-suite run. It was confirmed **pre-existing** by running the suite on the base tree, where it appears identically. That file is not in this diff and the issue is out of scope here.

## Merge with main

`origin/main` advanced 14 commits during planning, three of which touched files this change edits (#1730's sidebar row chips and Telegram glyph, #1723's Ctrl+Shift+W). All three auto-merged with no conflict and nothing was hand-edited, which was verified rather than asserted:

```
git merge-tree --write-tree ec8b1344 7ef1af8d  ->  f6ae3de26fa6814d12ae7e03a172ce122185e17f
git rev-parse baf5ee10^{tree}                  ->  f6ae3de26fa6814d12ae7e03a172ce122185e17f
```

Tree OIDs are identical, so the merge is exactly git's unaided three-way resolution across the whole tree, not only the five paths in this diff.

The two changes are adjacent rather than interacting. `main`'s `sidebar.css` delta contains no `session-context*` selector; its new `.agent-name-chip` lives on `.replica-item` while this icon lives in the portalled context menu. None of main's new or changed test files references `session-context-option-icon`, `1F465`, `Add to Group` or `Create new group`. `variables.css` is untouched by main, so every contrast figure in the shipped comments is still correct.

## Disclosed contrast shortfall

| Theme | Icon | Surface | Ratio |
| --- | --- | --- | --- |
| Dark | `--status-waiting` `#22c55e` | `--sidebar-surface` `#0e0e18` | 8.42 : 1 |
| Light | `--status-waiting` `#16a34a` | `--sidebar-surface` `#ebebf0` | **2.77 : 1** |

The light value falls short of the 3:1 non-text minimum. This is disclosed and accepted, the same class of shortfall #1708 accepted for the light amber at 2.47. Retuning the light token is a legitimate follow-up; hard-coding a darker green at the call site is blocked by the new CSS test case.

#1731 and #1719 originally recorded this as 2.78 : 1. That was a rounding error of the same computation (2.77375531), caught during plan review, and both issues have been corrected with the derivation recorded.

## Acceptance criterion 7

AC 7 asks for visual verification in the built app in both themes, with `1731-v6-dark.png` and `1731-v6-light.png` in this PR body.

**Those screenshots are not attached.** The user reviewed the change and signed it off as good, and directed that it land. Recording that plainly rather than presenting it as the screenshot evidence the criterion asks for: the visual check has the user's approval, not the artefacts.

A testable binary built from this merge tip is at `target/release/agentscommander_testeable.exe`, SHA-256 `45bc5c9aa691a2d26f506b5937cbecccba75b3933fd828589bb5c76b670bf1c2`, if anyone wants to look at it later.

Closes #1731

https://claude.ai/code/session_01VAWfjheWKJaSsqYK3ej7fs


mblua:
## Merged with `--admin` over a red `rust-regression`, and why

`rust-regression` (Windows) is red on this PR. It is red for a reason that has nothing to do with this change, and it is red on `main` too. Recording the reasoning here rather than in a commit message, so the decision is auditable.

**This change contains no Rust.**

- `git diff --name-only origin/main..HEAD | grep -c '\.rs$'` is `0`
- `git rev-parse HEAD:src-tauri` = `git rev-parse origin/main:src-tauri` = `8fe4ee40313805f1fa40001d2f687a34ae7f1f76`
- `Cargo.lock` unchanged

The Rust that `cargo test` compiles and runs on this branch is bit-for-bit what is on `main`, so this branch cannot have caused the failure.

**The failure is a 15 second timeout, not an assertion:**

```
thread 'agent_version::tests::probe_version_skips_an_ansi_only_first_line' panicked at src-tauri\src\agent_version.rs:1417:9:
  left: Failed("timed out after 15s (killed)")
```

3979 pass, 1 fails. `rust-regression-linux` and `rust-regression-macos` pass on the same commits; the Unix branch of that test spawns `sh`, the Windows branch pays PowerShell's cold start against a 15 second budget.

**Same bytes, both outcomes:**

| When (UTC-3) | Commit | Result |
| --- | --- | --- |
| 01:09 | `1acd528e` | pass |
| 03:56 | `b8fc5bfe` | pass |
| 20:17 + 20:18 | `baf5ee10` | fail (push/PR twins, one minute apart — one observation, not two) |
| 20:46 | `baf5ee10` | fail (independent rerun, half an hour later) |

**The check that exercises this change passes.** `frontend-regression` is green on both runs, in 1m42s and 1m43s.

The gate breakage is tracked in #1741 and the evidence is also recorded on the known-flaky inventory #1241. It blocks every PR opened right now, not only this one; it was not introduced here, and fixing it is not in this issue's scope.

Issue #1743, opened 2026-09-04
test(#1737): give the two CLI loaders a path-taking seam and write the four test obligations the plan could not satisfy


mblua:
Follow-up to #1737. Carries the erratum recorded there (PL-1) to closure.

## Situation

The #1737 plan (`plans/1737-local-config-overlay.md`, `Plan-SHA256: 168BD2A74B09EA7306E74369BF1C7AADCE24190085273FCAEEAE8144CC3D3A44`) mandates four test obligations that name the two CLI loaders, and simultaneously closes every mechanism that could satisfy them. The inconsistency was in the frozen bytes and was certified by all three planning parties across four consensus rounds; it surfaced only at implementation.

The unimplemented obligations:

| Id | Obligation |
| --- | --- |
| S24 | `load_settings_for_cli_strict` **on a path** with a valid overlay returns the merged view |
| S20, second half | ...and for `load_settings_for_cli_strict` with an absent base |
| S30, last sentence | The same four rejection arms asserted **through `load_settings_for_cli_strict`** |
| S31 | The four B-reverse edges, one test per row of 2.11b, **each through all three loaders** (only `load_settings_from_path` is driven today) |

`AC-17`'s evidence column cites S31 for "in all three loaders", so an acceptance criterion is pinned to evidence the plan makes unproducible.

## Why the obligations could not be met

Neither `load_settings_for_cli` nor `load_settings_for_cli_strict` takes a path. Both resolve through `settings_path()` (`src-tauri/src/config/settings.rs:1945-1947`, private) to `config_dir()` (`src-tauri/src/config/mod.rs:911-913`) to `instance_location()`, a process-wide `OnceLock` resolved once at first call. A probe run during implementation confirmed the redirect does not take effect:

```
before=Some(".../debug/deps/.agentscommander_lib-a5dca853102a9c29")
after =Some(".../debug/deps/.agentscommander_lib-a5dca853102a9c29")
requested="C:\Users\maria\AppData\Local\Temp\.tmpVGYHVc"
redirect_took_effect=false
```

[...]

Issue #1710, opened 2026-09-02
Flaky Windows loopback WebSocket shutdown test needs a deterministic teardown contract


mblua:
## Situation

While validating #1690 at commit `8cd555f3a04983c695694fbded34e0b2f90fa570`, the required Windows Rust library harness intermittently failed:

```text
tests::real_loopback_websocket_is_closed_and_reaped_by_generation_stop
read client EOF: Os { code: 10054, kind: ConnectionReset,
message: "An existing connection was forcibly closed by the remote host." }
```

This is not caused by #1690. The only #1690 change in `src-tauri/src/lib.rs` is two Tauri invoke-handler registrations; the loopback test app installs no invoke handler, sends no application frame, and does not execute the changed agent-update routes.

## Cause

The test proves server-task completion and admission draining, then overconstrains the client's transport observation to orderly TCP EOF (`Ok(0)`). The generation-cancellation path drops the WebSocket halves without first completing an RFC 6455 Close handshake. On Windows, that same successfully reaped lifecycle may surface as `WSAECONNRESET` / error 10054.

The mandatory Windows PR job runs `cargo test --lib --bins --tests` with normal parallelism and no retry, so this pre-existing teardown-contract mismatch can block unrelated pull requests.

## Required decision and change

Choose the intended product contract before implementation:

1. If the requirement is generation stop plus complete task/admission reaping, keep those lifecycle proofs and accept either orderly EOF or Windows connection reset after they succeed.
2. If the product requires a graceful RFC 6455 shutdown, update the production WebSocket path to send and complete a Close handshake before teardown, then require that stronger behavior in the test.

Do not weaken lifecycle assertions merely to accept an early or unreaped disconnect.

## Acceptance criteria

- The selected shutdown contract is documented and enforced deterministically.
- Server/read tasks are joined and the admission count drains before the client teardown result is accepted.
- Repeated Windows runs use a predetermined count and all attempts pass; all attempts are reported.
- `cargo test --lib --bins --tests` passes on Windows without retry.
- No WebSocket task leak, premature admission release, or generation-stop regression is introduced.

## Traceability

- Discovered during the external Phase-02 gate for #1690 under epic #1672.
- Deferred by explicit user-approved Phase-02 exception; any reproduction in required PR CI or Phase 05 blocks delivery until resolved.
- Related flake inventory: #1039.

Issue #1741, opened 2026-09-04
ci(rust-regression): the Windows gate fails reproducibly on main, blocking every PR


mblua:
## Situation

`rust-regression` (the Windows leg) fails reproducibly on code that is already on `main`. It is a required check on the `main` ruleset, so **every pull request opened right now is blocked by it**, regardless of what that pull request changes.

A single test fails:

```
thread 'agent_version::tests::probe_version_skips_an_ansi_only_first_line' panicked at src-tauri\src\agent_version.rs:1417:9:
assertion `left == right` failed
  left: Failed("timed out after 15s (killed)")
```

3979 tests pass, 1 fails, in every observation.

## The failure is a timeout, not a behavioural assertion

`src-tauri/src/agent_version.rs:1401-1419` spawns a real process and waits for it to print a version string, with `PROBE_TIMEOUT` at 15 seconds:

- Windows: `powershell.exe -NoProfile -NonInteractive -Command "[char]27 + '[?25l'; '2.0.1'"`
- Unix: `sh -c "printf '\033[?25l\n2.0.1\n'"`

The test's purpose is to assert that an ANSI-only first line is skipped. What it actually measures on a loaded runner is process startup latency. `rust-regression-linux` and `rust-regression-macos` pass on the same commits, which fits: `sh` starts in milliseconds, PowerShell's cold start is charged against a 15 second budget.

## Same Rust bytes, both outcomes

The `src-tauri` tree OID is `8fe4ee40313805f1fa40001d2f687a34ae7f1f76` at every commit below, and `agent_version.rs` is unchanged across all of them, so nothing about the code under test differs between a pass and a fail.

| When (UTC-3) | Commit | Branch | Result |
| --- | --- | --- | --- |
| 2026-09-03 01:09 | `1acd528e` | feature/1723 | pass |
| 2026-09-03 03:56 | `b8fc5bfe` | feature/1730 | pass |
| 2026-09-03 20:17 | `baf5ee10` | feature/1731 | **fail** |
| 2026-09-03 20:18 | `baf5ee10` | feature/1731 | **fail** |
| 2026-09-03 20:46 | `baf5ee10` | feature/1731 | **fail** (rerun) |

[...]

Issue #1726, opened 2026-09-02
Flaky on the Windows runner: probe_version_skips_an_ansi_only_first_line and neighbouring wall-clock tests


mblua:
`rust-regression` (the Windows leg, the only one that runs `cargo test`) fails intermittently on wall-clock tests against real process spawns. It is not tied to any one branch.

## Evidence

`agent_version::tests::probe_version_skips_an_ansi_only_first_line` (`src-tauri/src/agent_version.rs:1417`) failed on:

- `main` itself, commit `48783767`, job 100131551664, 05:20Z
- `feature/1691-frontend-cancellation-and-outcomes`, commit `d52f8738`, 13:51Z
- `feat/1682-terminal-last-agent-message-timestamp`, four separate runs, 13:00Z to 14:02Z

The observed failure mode is a timeout, not a wrong value:

```
assertion `left == right` failed
  left: Failed("timed out after 15s (killed)")
 right: Version("2.0.1")
```

The test spawns `powershell.exe` and gives it 15 s. Under runner load that is not enough.

Two neighbouring tests failed in the same window with the same shape, and only in some runs:

- `session::selection::tests::capped_shutdown_abort_joins_pending_container_stop_exactly_once` (`selection.rs:3414`)
- `session::selection::tests::capped_shutdown_abort_joins_real_pending_container_start_and_stop_exactly_once` (`selection.rs:3749`)
- `pty::spawn_diagnostics::tests::monitor_writes_the_startup_verdict_for_a_silent_child_and_keeps_watching` (`spawn_diagnostics.rs:1953`)

Five runs of one identical commit produced four different failing sets, and one run passed tests that its same-SHA twin failed. That is the signature of an environment-sensitive timeout, not a regression.

Measured rate over recent `main` commits: 18 success, 1 failure, 5 cancelled.

## Suggested direction

Raise `PROBE_TIMEOUT` for the CI environment, or make these tests tolerant of spawn latency rather than pinning a wall-clock budget that a loaded shared runner cannot honour. The cluster of failures in one time window suggests runner-pool degradation rather than a steady rate, so a fixed higher bound may be enough.

Found while landing #1682, whose PR #1716 was merged with admin over this red check with the control evidence recorded in the PR.

PR #1744, opened 2026-09-04
feat(config): .local alter-ego override layer for context templates and instance settings (refs #1737)

### LINKED ISSUE #1737: config: support a .local alter-ego override layer for context templates and instance settings
## Situation

AgentsCommander has no "alter ego" override layer for its own configuration. Today the only `*.local.json` files in the repository belong to Claude Code's own convention, not to AC:

- `.claude/settings.local.json` (repo root, 5 lines)
- `docs/integrations/rtk_claude/settings.local.json` (49 lines), the tracked mirror the seeder copies to `<workspace>/.ac/default.claude/settings.local.json`

AC provides those files to the coding agents it seeds, but AC itself never reads a local variant of anything. A grep for `overlay`, `config_override` and `merge_config` over `src-tauri/src/config/` returns nothing. `local_config_io.rs` is a false positive: it is the atomic JSON object writer (`update_config_json_object`, mutex + temp file + rename), not an override layer.

The only alter-ego mechanism AC has today works at the directory level, not the file level. `config_dir()` (`src-tauri/src/config/mod.rs:905-912`) derives the instance config directory from the binary file stem: `C:\tools\agentscommander_ac2.exe` resolves to `C:\tools\.agentscommander_ac2\`. Running a parallel instance therefore requires renaming the executable, and there is no way to keep one base configuration while overriding a handful of values for a specific machine, instance, or debugging session.

## Why the change is needed

An operator running several AC instances, or debugging one, needs to change a small number of values without editing the shared base file:

- keep `settings.json` under version-controlled or synchronized state while one instance overrides `logLevel`, a bind address, or a watcher interval
- try a modified coordinator context without destroying the seeded `Context.coordinator.md`, which AC auto-upgrades and records in the seed manifest

Today both require editing the base file in place. The base value is then lost, and for the context templates the edit collides with the auto-upgrade and recognizer machinery in `seeded_context_templates.rs`.

## The trap this issue must avoid

`settings.json` is not a user-owned file. AC rewrites it on its own: `load_settings_from_path` (`settings.rs:1870`) auto-generates `root_token` when missing, runs migrations and sets `needs_save`; there are 7 callers of `save_settings()` (`settings.rs:4274`), and the UI persists the whole `AppSettings` struct whenever a window moves, the zoom changes, or a project is opened.

A naive in-memory merge is therefore actively harmful. Sequence:

1. `settings.json` holds `"logLevel": "info"`
2. the operator writes `settings.local.json` with `"logLevel": "debug"`
3. AC loads and merges; in memory `logLevel` is `"debug"`, which is correct
4. the operator moves a window; AC serializes the in-memory `AppSettings` and writes `settings.json`
5. `settings.json` now holds `"debug"`; the original `"info"` is gone
6. deleting `settings.local.json` no longer restores `"info"`

The override is silently absorbed into the base file on the first save, and the damage is invisible until the operator tries to turn the override off.

## How it will be changed

Two distinct behaviours, one per file type.

### Markdown: whole-file substitution

When `<name>.local.md` exists next to `<name>.md`, the local file is used and the base file is not read.

Injection point: `read_or_create_context_recorded`, reached from `session_context.rs:2209-2219`. The same helper serves all three context templates, so all three gain the override:

| Constant | Value | Declared at |
| --- | --- | --- |
| `COORDINATOR_CONTEXT_TEMPLATE_FILENAME` | `Context.coordinator.md` | `session_context.rs:13` |
| `GLOBAL_CONTEXT_TEMPLATE_FILENAME` | `Context.AgentsCommander.md` | `instance_artifacts.rs:158` |
| `ROOT_AGENT_CONTEXT_TEMPLATE_FILENAME` | `Context.root-agent.md` | `instance_artifacts.rs:155` |

The `.local.md` file is a pure read-time override and is fully operator-owned. AC must never create it, never sync it, never auto-upgrade it, never retire it to a `.retired-*.bak`, and never record it in the seed manifest. The existing recognizer, migration and `read_or_create` behaviour for the base `.md` stays exactly as it is today, including creating the base file when it is absent.

### JSON: deep merge, the final key wins

When `settings.local.json` exists next to `settings.json`, the two are merged in memory once, at load, and the merged value is what the rest of the process reads.

The merge is recursive, not a first-level replace. The override applies only at the leaf key:

```
settings.json        { "watchers": { "a": { "enabled": true, "intervalMs": 500 } } }
settings.local.json  { "watchers": { "a": { "intervalMs": 50 } } }
effective            { "watchers": { "a": { "enabled": true, "intervalMs": 50 } } }
```

`enabled` survives because `watchers.a` is an intermediate node, not a final key. This matters in practice: `AppSettings` (`settings.rs:312`, 77 fields) nests through `coding_agent_profiles`, `main_geometry`, `watchers: BTreeMap<String, WatcherEntry>`, `auto_self_clear_by_agent` and `agent_auto_update_by_command`.

Decided semantics:

1. **Arrays are leaves.** An array in the local file replaces the base array whole; there is no element-wise or index-wise merging. `[]` in the local file therefore empties the list. This keeps the override predictable and matches Claude Code, VS Code and tsconfig.

2. **The base file is preserved on write.** At load, AC captures both the local overlay and the original base value of every leaf the overlay touches. On save, AC serializes the current `AppSettings` and restores those leaves to their captured base values before writing `settings.json`. Everything the overlay does not touch saves normally.

   ```
   settings.json        { "logLevel": "info",  "mainZoom": 1.0 }
   settings.local.json  { "logLevel": "debug" }
   operator moves the window, AC saves
   settings.json        { "logLevel": "info",  "mainZoom": 1.4 }
   in memory            logLevel = "debug"
   ```

   Deleting `settings.local.json` restores exactly what was there before. The accepted cost: a UI edit to a key the overlay owns is silently discarded. Surfacing that in the UI is deliberately out of scope here and belongs in a follow-up.

3. **Merged once, cached.** The merge happens at load and the result is held in memory; it is not recomputed per value read.

4. **Absent or malformed local file.** A missing `settings.local.json` is the normal case and changes nothing. A malformed one must not silently fall back to base only; the existing parse-failure behaviour for `settings.json` (`log::error!` then defaults) is the precedent to follow deliberately, and the plan must state the chosen behaviour explicitly.

### Readers that would bypass the overlay

These must be reviewed and either routed through the merged view or explicitly documented as out of scope:

- `load_settings_for_cli()` (`settings.rs:2019`) and `load_settings_for_cli_strict()` (`settings.rs:2090`)
- `cli/terminal_snapshot.rs:249`, which reads the raw `settings.json` bytes through `read_bounded_regular` to extract the token

### Also to settle in the plan

- whether `settings.local.json` and `*.local.md` must be added to the managed-artifact inventories in `instance_gitignore.rs` and `instance_artifacts.rs`, or deliberately left out as operator-owned files

## Acceptance criteria

1. With `Context.coordinator.local.md` present, a coordinator session receives its bytes and not those of `Context.coordinator.md`; with it absent, behaviour is byte-identical to today. Same for `Context.AgentsCommander.md` and `Context.root-agent.md`.
2. AC never creates, syncs, upgrades, retires or manifest-records any `*.local.md`.
3. A nested override in `settings.local.json` replaces only the final key and leaves sibling keys from `settings.json` intact.
4. An array in `settings.local.json` replaces the base array whole, and `[]` yields an empty list.
5. After AC saves `settings.json` for an unrelated reason, every leaf owned by `settings.local.json` still holds its original base value on disk, and unrelated changes are persisted.
6. Deleting `settings.local.json` restores the exact pre-override effective configuration.
7. A missing `settings.local.json` produces behaviour byte-identical to today.

mblua:
Closes #1737

Adds a `.local` alter-ego override layer so an operator can keep a base configuration and override a few values per instance or per machine, without the app absorbing the override into the base file.

## What it does

**Markdown, whole-file substitution.** When `<name>.local.md` exists next to `<name>.md`, the local file is used and the base is not read. Covers all three context templates: `Context.coordinator.md`, `Context.AgentsCommander.md` and `Context.root-agent.md`. The `.local.md` is operator-owned: AC never creates, syncs, auto-upgrades, retires to `.retired-*.bak` or seed-manifest-records one, and the base file's recognizer, migration, sync and self-healing behaviour is unchanged.

**JSON, deep merge with the final key winning.** When `settings.local.json` exists next to `settings.json`, the two are merged in memory once at load. The merge is recursive and the override applies only at the leaf, so a nested override does not wipe its siblings:

```
settings.json        { "watchers": { "a": { "enabled": true, "intervalMs": 500 } } }
settings.local.json  { "watchers": { "a": { "intervalMs": 50 } } }
effective            { "watchers": { "a": { "enabled": true, "intervalMs": 50 } } }
```

Decided semantics:

- **Arrays are leaves.** An array in the local file replaces the base array whole; `[]` empties the list.
- **The base file is preserved on write.** At load, the base value of every leaf the overlay owns is captured; on save, those leaves are restored before `settings.json` is written. Everything the overlay does not touch saves normally. Deleting `settings.local.json` restores the exact pre-override configuration.
- **Merged once at load and cached**, not recomputed per read.

The base-preservation half is the load-bearing part. `settings.json` is not a user-owned file: the app auto-generates `rootToken`, runs migrations, and persists the whole `AppSettings` whenever a window moves or a project opens. Without it, the first such save copies the override into the base permanently and silently, and deleting the local file no longer restores anything.

## Files

Eight paths, one owner (Rust), one contract surface. No IPC signature, wire payload, CLI verb or on-disk schema change.

| Path | Change |
| --- | --- |
| `src-tauri/src/config/local_overlay.rs` | new leaf module: typed diagnostics, Markdown override reader, JSON deep merge, restore-plan capture with the derived-id closure, restore and re-apply |
| `src-tauri/src/config/mod.rs` | one `pub(crate) mod local_overlay;` |
| `src-tauri/src/config/settings.rs` | hidden overlay field, five key tables, the merge, seven default arms, the merged top-level reader behind the two raw readers, three writer insertions, the CAS re-apply and carry-over, ten migration suppressions |
| `src-tauri/src/config/session_context.rs` | four private helpers, three substitution points |
| `src-tauri/src/commands/config.rs` | two `local_overlay_state` carry-overs |
| `src-tauri/src/config/instance_artifacts.rs` | two constants, two `Ignore` rows so `settings.local.json` and `*.local.md` are not committed |
| `src-tauri/src/config/instance_gitignore.rs` | tests only |
| `src-tauri/module-arcs.txt` | regenerated |

## Review

Planning ran four consensus rounds before any code was written. `ac-architect-v4` authored the plan; `ac-dev-rust-v4` and `ac-dev-rust-grinch-v4` reviewed each candidate independently and in parallel. Rounds 1 to 3 were rejected. Defects the reviewers found and closed before implementation include:

- the renderer round-trip erasing the overlay state, so the first window move wrote the override into the base file, the exact trap the issue exists to prevent;
- `logLevel`, the issue's own headline example, bypassing the overlay entirely through a raw reader;
- UI edits to coding-agent profiles being silently discarded for the whole session;
- any key whose field carries `skip_serializing_if` reverting to the base value in memory on the first save;
- a legacy migration overwriting an eligible overlay-owned key;
- three inventory completeness claims that could not be backed, and several acceptance criteria that could not go red.

Round 4 reached three-party consensus on `Plan-SHA256: 168BD2A74B09EA7306E74369BF1C7AADCE24190085273FCAEEAE8144CC3D3A44`, which was then approved by the user at the human review gate before implementation began.

Implementation review found no defect in the shipped production code and no deviation from the plan that changes behaviour. Two non-blocking observations were fixed (`14d07b4d`, `7379a1a4`) and re-reviewed. The Step 9.5 merge of `origin/main` was verified to leave the whole `src-tauri` tree byte-identical (one recursive tree hash), so the code clearance carries forward unchanged.

## Verification

On the merged tree, from the repository root:

| Command | Result |
| --- | --- |
| `cargo fmt --all -- --check` | exit 0 |
| `cargo clippy --workspace --all-targets -- -D warnings` | exit 0 |
| `cargo test --lib --bins --tests` | exit 0 |

Lib binary, read directly from the harness output rather than derived: header `running 4082 tests`, result `4062 passed; 0 failed; 20 ignored`. The change adds 82 tests (32 `local_overlay`, 32 `settings`, 12 `session_context`, 6 `commands/config`), none gated on `cfg(unix)`, so all of them execute on the Windows leg.

No whole-suite aggregate is quoted here deliberately: `pty_input_cross_process` re-invokes itself with a filter and its child output interleaves into the parent's captured stream, so both the sum of all `test result:` lines and the first-line-per-executable sum are line-attribution artifacts rather than measurements.

**Dependency cycle gate.** Two new arcs, both into a new zero-out-degree leaf module, so both are cycle-neutral regardless of the source module's SCC:

```
+agentscommander_lib::config::session_context -> agentscommander_lib::config::local_overlay
+agentscommander_lib::config::settings        -> agentscommander_lib::config::local_overlay
```

`module-arcs.txt` gains exactly those two lines and nothing else. The new module imports only `std::path` and `serde_json`; its production half contains zero occurrences of `crate::`, `super::` or `log::`.

## Known gap, recorded before landing

The frozen plan mandates four test obligations that name the two CLI loaders and simultaneously closes every mechanism that could satisfy them: neither loader takes a path, both resolve through a process-wide `OnceLock`, section 4.2i refuses to refactor them into a shared helper, and the env-var route is both unacceptable and ineffective. The inconsistency was inside the certified bytes and surfaced only at implementation.

Production code is complete and exactly as mandated on the CLI paths. The genuinely untested surface is six inline conjuncts, which review extracted mechanically and found identical to six conjuncts that are tested. Neither CLI loader saves, so the worst case is an in-memory-only, non-destructive miss on the non-GUI load path; the disk-preservation direction is untouched.

Shipped with the gap stated at the user's explicit decision. Recorded as an erratum on #1737, with the fix filed as #1743.

## CI note

`real_loopback_websocket_is_closed_and_reaped_by_generation_stop` may go red on the required Windows leg. That is #1710, a documented pre-existing flake, not this change: the merge altered no Rust byte, `lib.rs` is not among these eight paths, and the test passed 5 of 5 isolated runs and on a second full-suite run locally.


mblua:
## Merging with `--admin` over a red required check. Justification, recorded before the merge.

`rust-regression` is a required status check on the `main` ruleset and is red on this PR. Merging with `--admin` bypasses it. That is a deliberate decision taken with explicit user authorization, not a routine admin merge, so the reasoning is recorded here.

### The failure is #1741, not this change

Both failing attempts fail on the same test, in a file this PR does not touch:

```
thread 'agent_version::tests::probe_version_skips_an_ansi_only_first_line'
panicked at src-tauri\src\agent_version.rs:1417:9:
assertion `left == right` failed
  left: Failed("timed out after 15s (killed)")
```

#1741, opened before this PR, documents exactly this and states: "`rust-regression` (the Windows leg) fails reproducibly on code that is already on `main`. It is a required check on the `main` ruleset, so every pull request opened right now is blocked by it, regardless of what that pull request changes."

Its diagnosis is that the test spawns a real process and measures PowerShell cold-start latency against a 15 second `PROBE_TIMEOUT` (`agent_version.rs:1401-1419`), which is why the Linux and macOS legs pass on the same commits.

### Evidence that this PR is not the cause

1. **The same check passed on the identical head.** Both workflow runs are on `a0e8cb12543d0903e277a5de184fa559cc897b65`; the `pull_request`-event run reports `rust-regression` **pass** (17m56s), the `push`-event run reports fail. Same head SHA, opposite results.
2. **`agent_version.rs` is not among the eight changed paths.** `git diff --name-only 302e6c12 HEAD` lists only `module-arcs.txt` and seven files under `src-tauri/src/config/` and `src-tauri/src/commands/config.rs`.
3. **Unrelated branches fail the same way.** `feature/1742-release-tag-authorisation` and `feature/1731-add-to-group-user-plus-icon` both failed this gate on their own commits, and #1741 tabulates the same `src-tauri` tree OID producing both outcomes across several branches.
4. **Every other required check is green**, including the two other Rust legs and the Windows one that actually exercises this change: `rust-regression-linux`, `rust-regression-macos`, `windows-release-cli-smoke`, `terminal-snapshot-portable` on all four platforms, `rust-fmt`, `test-debt`, `frontend-regression`, `validate-branch-name`, `lockfile-drift`.
5. **The local gate is green on this exact tree**: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings` and `cargo test --lib --bins --tests` all exit 0, with the lib binary at `4062 passed; 0 failed; 20 ignored`.

A hypothesis that this PR's 82 new tests added enough wall-clock load to tip a timing-sensitive test over its budget was considered and is refuted by point 3: the same failure reproduces on `main` and on branches that add no tests at all.

### What this does not claim

It does not claim the gate is healthy. #1741 remains open and blocks every pull request in this repository; #1713 and #1726 track the same test from the flake angle, and #1710 tracks a second Windows flake. Unblocking that gate is worth prioritising on its own, separately from this delivery.

Issue #1833, opened 2026-09-06
P5 - Document the two selected-row rail settings


mblua:
Phase **P5** of epic #1796. Class `mechanical`, owner docs, 1 file, depends on P1, P3 and P4b.

**Specification:** `plans/1796-selected-row-rail-setting/p5-docs-settings-reference.md`
**Phase-SHA256:** `056426814D2B24B1689741AD71C98AB56E3789E22B3BF039B4385C8F1AFDAEB5` (19832 bytes, 317 lines)

Verify that digest before touching anything, then read that file in full. It is the sole specification for this phase.

## Situation

The two new settings are undocumented. `docs/reference/settings.md` documents every other field.

## Files

| Path | Change |
|---|---|
| `docs/reference/settings.md` | two table rows, inserted after the `sidebarStyle` row at `:261` |

The two rows are given **verbatim** in the phase file at `:50` and `:51`. Copy them; do not compose prose.

## Acceptance criteria

Ten numbered criteria, authoritative in the phase file. **Criterion 7 is the one that matters**: a `sha256sum` of the two inserted rows must equal `4e65de7b05450910ec14c7b74d826d3d8e7ce691a54fa91bf2ea01c78e00b895` at **922 bytes**. Six contradictory or truncated variants of those rows pass every other criterion; only this digest separates them, and byte count alone does not, since two different variants both measure 912 bytes.

`tr -d '\r'` in that pipeline is load-bearing: the target file is CRLF and the plan fence is LF.

Criterion 5 pins the documented formats to the code on three sides: `WIDTH_RE`, `RAIL_WIDTH_MAX` and `COLOR_RE` in `src/sidebar/selected-row-rail.ts`, and the Rust doc comment in `src-tauri/src/config/settings.rs`. Criterion 6 requires `grep -c -i obsidian` to print `1`, the pre-existing `sidebarStyle` row, with the match quoted in the report.

## Verification

The phase's own verification block, then `npm test`, which does not validate the prose but proves this phase changed nothing else. No CI check lints `docs/`.

**Every command runs from a script file**, named in the report. Four of the seven patterns expect `0`, where a false zero is a false **pass**. Do not run any of them standalone: that form was measured producing a stably wrong digest twice in one session.

Run **after** P4b has reported, not alongside it: while a CTL-6 mutation is in place the CSS suite is genuinely red, and that failure belongs to P4b, not to this phase.

Issue #1832, opened 2026-09-06
P4b - Prove the selected-row rail resolves as designed


mblua:
Phase **P4b** of epic #1796. Class `design-bearing`, owner frontend, 1 file, depends on P4a and P3. Runs before P5.

**Specification:** `plans/1796-selected-row-rail-setting/p4b-css-rail-proof.md`
**Phase-SHA256:** `E91DE2369523ACD5A7A115CB1EE9E5EBBEA8AC23A031F4BE9F51E8134DB1B0AD` (24148 bytes, 400 lines)

Verify that digest before touching code, then read that file in full. It is the sole specification for this phase.

## Situation

P4a's 36 edits are asserted by nothing. A CSS-text assertion that silently matches zero rules looks exactly like a passing one, so the proof has to measure rather than assert.

## Files

| Path | Change |
|---|---|
| `src/sidebar/styles/selected-row-rail-css.test.ts` | **new** |

## What it proves

A 7 styles x {dark, light} x {session, replica} matrix, 28 cells, resolved in jsdom by building a real DOM, matching all 44 censused rules, expanding every border shorthand into its left longhands in declaration order, and picking the winner by (specificity, document order). Plus a six-cell coordinator ledger and an 18-entry preserve ledger.

## Controls, all load-bearing

CTL-1 pins the census at `83 / 44 / 51 / 55 / 11` with the eleven decoys named. CTL-2 requires every extraction to throw on a miss and every ledger selector to resolve to exactly one rule. CTL-3 gates jsdom's `:has()` support before any cell is built. CTL-5c, CTL-5d and CTL-5e cover `:has()` specificity, `:not(:has())` and equal-specificity document order. CTL-6 runs two recorded mutations that must each fail at their target rather than at an upstream control. A8 asserts that the count of contexts moving at any accepted width is **zero**, so raising the cap to 15 makes it 3 and fails.

**CTL-6 mutates `sidebar.css`, which this phase does not own.** P4a must be committed and `git status --porcelain -- src/sidebar/styles/sidebar.css` must be empty before either mutation. Restore with `git checkout --` only: `git show HEAD:<path> > <path>` prints the LF blob into a CRLF worktree and silently rewrites every line ending.

## Verification

`npm test`. The phase report must state that rendered geometry is unproven by this suite; that belongs to the Step 7.5 human gate in `epic.md` section 14.

Issue #1831, opened 2026-09-06
P4a - Route the selected-row rail through three custom properties


mblua:
Phase **P4a** of epic #1796. Class `design-bearing`, owner frontend, 3 files, depends on nothing, parallel with P1, P2 and P3.

**Specification:** `plans/1796-selected-row-rail-setting/p4a-css-rail-tokens.md`
**Phase-SHA256:** `B23DC24C967F257CFD929AE7305A7E1B782F3A038202BEB3849C692418ADF1F8` (23181 bytes, 400 lines)

Verify that digest before touching code, then read that file in full. It is the sole specification for this phase.

## Situation

The selected-row rail's width and colour are hardcoded in `sidebar.css` and overridden per theme, so a setting cannot reach them. `--sidebar-accent` cannot be retargeted: it is read across the file and paints far more than this rail.

## Files

| Path | Change |
|---|---|
| `src/sidebar/styles/variables.css` | three new custom properties, in the `:root` block **only**, never in `html.light-theme` |
| `src/sidebar/styles/sidebar.css` | **36 rule blocks** and one comment |
| `src/sidebar/styles/working-tint-css.test.ts` | one assertion |

## The mechanism

26 CHANGE rules are re-pointed **in place**; 18 PRESERVE rules keep their declarations. `26 + 18 = 44`. A third property, `--ac-rail-delta`, is resolved per element by the same cascade that used to decide the width, and every row rule declaring a left padding is rewritten as `max(0px, calc(<original> - var(--ac-rail-delta)))`, so `padding-left + border-left-width` is unchanged everywhere.

**Rules are identified by full selector text, never by line number.** Rule 36's selector resolves to two byte-identical blocks: take the one declaring the `padding` shorthand; the other declares only `color` and is not touched. Rule 39 is load-bearing: its `--ac-rail-delta: 0px` is the only thing keeping the excluded `obsidian-mesh` coordinator row from compensating for a rail that never scales.

Two behavioural exceptions stay byte-identical: the `deep-space` coordinator card and the `obsidian-mesh` coordinator row.

## Acceptance criteria

Authoritative in the phase file. The prescribed `grep -c -- "var(--sidebar-accent)"` must print **98**; a printed 99 means rule 13, the selected replica row in `noir-minimal`, was missed.

## Verification

`npm test`, plus the phase's greps run from a script file and named in the report.

Issue #1829, opened 2026-09-06
P2 - Mirror the two rail fields in the TypeScript contract


mblua:
Phase **P2** of epic #1796. Class `mechanical`, owner frontend, 8 files, depends on P1, parallel with P4a.

**Specification:** `plans/1796-selected-row-rail-setting/p2-typescript-contract.md`
**Phase-SHA256:** `8C915417D2A40246B40CDDE11EAAD7413B3999278D9A933DA8D4DD2922805E30` (6662 bytes, 148 lines)

Verify that digest before touching code, then read that file in full. It is the sole specification for this phase.

## Situation

`src/shared/types.ts` mirrors the Rust `AppSettings` by hand. P1 adds two always-serialized fields, so the mirror must gain them or the contract drifts silently.

## Files

| Path | Change |
|---|---|
| `src/shared/types.ts` | Two **required** fields beside `sidebarStyle` at `:654` |
| `src/shared/testing/ui-harness.tsx` | fixture literal, `:151` |
| `src/sidebar/components/AgentPickerModal.test.tsx` | fixture literal, `:66` |
| `src/sidebar/components/CodingAgentQuickConfiguration.test.ts` | fixture literal, `:48` |
| `src/sidebar/components/OnboardingModal.test.ts` | fixture literal, `:49` |
| `src/sidebar/components/settings-save.test.ts` | fixture literal, `:24` |
| `src/sidebar/components/SettingsModal.automation.test.ts` | fixture literal, `:100` |
| `src/sidebar/components/SettingsModal.test.ts` | fixture literal, `:10` |

The fields are **required, not optional**: 67 of the 73 mirror fields are required, the adjacent `sidebarStyle` is one, and the two new keys are always on the wire because they carry no `skip_serializing_if` and P1 forbids adding one.

## Acceptance criteria

Authoritative in the phase file. The positive control is load-bearing: run `npm run typecheck` **after the interface edit and before touching any fixture**, and it must name exactly those seven files, no more and no fewer. Both failure directions are handled explicitly.

## Verification

`npm run typecheck`, then `npm test` against the declared accepted-debt list, then the scoped `git status --porcelain` and `git diff --numstat` from the repository root. Counting commands run from a script file, named in the report.

PR #1838, opened 2026-09-06
feat(#1830): publish the selected-row rail settings to the DOM and edit them in the modal

### LINKED ISSUE #1830: P3 - Publish the rail settings to the DOM and edit them in the modal
Phase **P3** of epic #1796. Class `design-bearing`, owner frontend, 4 files, depends on P2, parallel with P4a.

**Specification:** `plans/1796-selected-row-rail-setting/p3-rail-application.md`
**Phase-SHA256:** `5DD328A757A8879954DD8BB5CAC43FD3B2CA2DA8B5F37BF38264A50DD64992D5` (20052 bytes, 400 lines)

Verify that digest before touching code, then read that file in full. It is the sole specification for this phase.

## Situation

Nothing publishes the persisted values to the DOM and nothing lets the user edit them. One leaf module owns both validators and the single writer, so the red hint and the decision to publish are the same rule.

## Files

| Path | Change |
|---|---|
| `src/sidebar/selected-row-rail.ts` | **new** leaf module: `WIDTH_RE`, `COLOR_RE`, `RAIL_WIDTH_MAX = 14`, the predicates and the writer |
| `src/sidebar/selected-row-rail.test.ts` | **new** |
| `src/sidebar/App.tsx` | one insertion at the existing `dataset.sidebarStyle` write |
| `src/sidebar/components/SettingsModal.tsx` | one import and one block in the Window section |

**Two writes are required**, not one: `App.tsx` runs once in async init and is not reactive, so the modal must write again inline on change, exactly as `sidebarStyle` does. `src/browser/` needs nothing, because the write sits outside the embedded guard.

## Decided behaviour, not open

Accepted width is 1 to 14 with an optional `px`; colour is `^#[0-9a-fA-F]{6}$`, three- and eight-digit forms rejected on purpose. Validation is a **soft warning** and never blocks the save. An invalid value is **not published**: the writer removes the custom property so `:root` supplies the factory `9px` / `#00ff5f`.

## Acceptance criteria

Authoritative in the phase file. Two stale-value controls, S1 for width and S2 for colour, **deliberately reuse one element across two calls** and override the file's own fresh-element rule; do not undo that. Two recorded mutations must each die at a bare `expect` in their own assertion.

The module is a new untracked file, so `git checkout --` cannot restore it and errors instead: the prescribed restore is a byte copy plus a matching `sha256sum`.

## Verification

`npm run typecheck`, `npm test`, and the phase's greps. Counting commands run from a script file, named in the report; several criteria expect `0`, where a false zero is a false **pass** rather than a false block.

mblua:
Phase P3 of epic #1796. **This is the phase that makes the feature visible.**

Closes #1830

## What lands here

P1 (#1828) persists the two settings, P2 (#1829) mirrors them in TypeScript and P4a (#1831) made the CSS read tokens. Nothing published those values to the DOM, and nothing let the user edit them. This does both.

Four files, **308 insertions, 0 deletions**:

| Path | Change |
|---|---|
| `src/sidebar/selected-row-rail.ts` | **new**, 74 lines. One leaf module owning `WIDTH_RE`, `COLOR_RE`, `RAIL_WIDTH_MAX = 14`, the two predicates and the single DOM writer |
| `src/sidebar/selected-row-rail.test.ts` | **new**, 175 lines, 69 tests |
| `src/sidebar/App.tsx` | +6, the startup call beside the existing `dataset.sidebarStyle` write |
| `src/sidebar/components/SettingsModal.tsx` | +53, the two fields and their red hints, between App Theme and the orchestrator wake checkbox |

After this, a selected sidebar row draws a **9px `#00ff5f`** rail, and **Selected Row Bar Width** and **Selected Row Bar Color** appear in Settings > General.

## Decided behaviour

Accepted width is **1 to 14**, optionally suffixed `px`; colour is `^#[0-9a-fA-F]{6}$`, three- and eight-digit forms rejected. 14 is the measured minimum `p + b` across the whole stylesheet, so at any accepted width **no row text moves**.

Validation is a **soft warning** and never blocks the save. An invalid value is **not published**: the writer removes the custom property, so `:root` supplies the factory `9px` / `#00ff5f` while the typed text persists as typed.

Two writes are required, not one: `App.tsx` runs once in async init and is not reactive, so the modal writes again inline on change, exactly as `sidebarStyle` does.

## Plan

`plans/1796-selected-row-rail-setting/p3-rail-application.md`, `Phase-SHA256: FB7550FD3FE52E92762E8972E7614F8CDF7AE02A6506D00732046F3B283DEA01`.

## Review

`ac-dev-rust-grinch-v4`: **PASS**, no defect, no frozen-plan change required.

The module is **byte-identical to the plan's fence**, proved by hashing both sides: the extracted LF fence and the committed blob agree at `d5c62af61779074ddfaf0ef37fb184fe52138afd8291572781e613de8aabe548`, and the CRLF worktree file at `8cc282dec63e7ef09cd4b4b11e537dd2a53d6be142a27d033a3eba6f89ce8e10`.

**Both prescribed mutations were run and each died at a bare `expect` in the assertion it was aimed at**, not at an upstream guard: M-P3-1 at `selected-row-rail.test.ts:138` (`'9px'` versus `''`), M-P3-2 at `:149` (`'#00ff5f'` versus `''`), each leaving 68 of 69 passing.

The reviewer then measured the counterfactual that justifies the design: **with the broken writer retained and only S2 neutralised, 69 of 69 pass.** S2 is therefore the only assertion that catches a writer leaving a stale colour on screen, which is why S1 and S2 deliberately reuse one element and override the file's own fresh-element rule.

The module is a new untracked file, so `git checkout --` could not restore it during mutation; it was restored from a `cmp`-verified byte copy and the digest is identical before mutating and after both restores.

## Verification

- `npm run typecheck`: exit 0
- `npm test`: exit 0, **178 files / 2153 tests**
- `grep -rc "applySelectedRowRail" src/browser/`: three files, all `:0`, exit 1. No second application point
- Writer call sites: 2 in `App.tsx`, 3 in `SettingsModal.tsx`
- The #1797 PtyAPI flake did not fire. A jsdom teardown line at `sessions.ts:58` is pre-existing and non-failing, confirmed by the reviewer having observed it before this phase and at `:52` before `main` shifted it

## Remaining phases

P4b (#1832) proves the CSS cascade resolves as designed; P5 (#1833) documents the two settings. Rendered geometry is unproven by any suite and belongs to the human gate in `epic.md` section 14.

https://claude.ai/code/session_018hqoFaCSK6PHQsea5LpEZ9


mblua:
## Landing with `--admin` over a red SonarCloud gate, on the repo owner's explicit instruction

Recording why, because the gate is red on purpose and not by oversight.

### What is red

`SonarCloud Code Analysis` reports **11.0% Duplication on New Code**, against a required ceiling of 3%. Every other check is green.

### The refactor that was supposed to fix it, and did not

The first diagnosis was that the duplication came from the two near-identical 26-line field blocks this PR added to `SettingsModal.tsx`. That diagnosis was **inference, not measurement**: it came from the arithmetic "11% of 308 new lines is roughly 34 lines, which is about that block", never from per-file data.

The plan was revised, the block was parameterised into a single `railField` render helper, and that landed here as `ad6b8734`. **Sonar then re-analysed `ad6b8734` and returned exactly 11.0% again, unchanged.** Removing about 26 duplicated lines did not move the metric by a decimal, which falsifies the original diagnosis.

The likely real source is `src/sidebar/selected-row-rail.test.ts`: 69 tests in 175 lines, where a five-line shape (`freshElement()`, the call, two `expect(...getPropertyValue(...)).toBe(...)`) repeats about a dozen times. A normalising scan of that file finds **22 repeated 6-line windows**, clustered around lines 92 to 120. That repetition is deliberate and is what makes each assertion independently killable; both prescribed mutations died at a bare `expect`, at `:138` and `:149`.

This is **not** confirmed against SonarCloud's own per-file numbers. The public API returns zeros for `new_duplicated_lines` under anonymous access, and the repository carries **no Sonar configuration at all**: no `sonar-project.properties`, no `.sonarcloud.properties`, and no Sonar reference in any workflow. Analysis scope, including whether test files are exempt from duplication, is configured in the SonarCloud UI, outside this repo.

### The other red check, and why it is not a defect

`rust-regression` failed once on the push-event run with two tests, `session::selection::tests::canceled_real_start_stop_error_retains_and_retries_the_deterministic_handle` and its `_panic_` twin, both panicking at `src-tauri/src/session/selection.rs:3772`.

It is a load flake, on three independent grounds:

1. The **same commit** `ad6b8734` passed that same job in the concurrent pull_request-event run.
2. This branch **touches no Rust at all**: four files, 307 insertions, 0 deletions, entirely under `src/sidebar/`.
3. The failing log reads `coordinator worker phase exhausted before shared deadline budget=200ms`, which is a timing budget, not a logic branch.

The job was re-run and **passed**.

### Human verification

The rendered geometry of this feature is unproven by any automated suite, which is why `epic.md` section 14 holds a human visual gate. A room-17 build was cut from this exact commit and inspected by the repo owner, who confirmed the result. That build:

- `agentscommander_standalone_room-17_20260907_021040.exe`
- SHA256 `f39a8ac365f863e3d6b3da46517a0bb489c6d7499bbb8a2c1e7303c69fbd69d6`, 62,563,840 bytes, version 0.30.5
- built from `ad6b8734d3f5b5bb089b7c5ac8c33a1472c934b3` with a clean worktree

### Decision

The repo owner was given three options, including lowering the gate, and chose to refactor first. After the refactor failed to move the metric, they instructed the merge to proceed with `--admin`.

**The gate is not being lowered and no Sonar setting is being changed by this merge.** The duplication question stays open against `selected-row-rail.test.ts` and needs authoritative per-file data from SonarCloud before anyone acts on it.

https://claude.ai/code/session_018hqoFaCSK6PHQsea5LpEZ9

Issue #1801, opened 2026-09-06
#1793 phase 1a: settings fields and the persisted-working predicate


mblua:
Child phase of epic #1793.

**Phase file (frozen):** `plans/1793-auto-resume-agents-on-restart/phase-1a-settings-and-predicates.md`
**Phase-SHA256:** `0971F581CF0CFD644C9795A8BE663BA7EFEF6A44CFB64146A17C2C4146086233`
**Epic Plan-SHA256:** `F11C050377EE0CA444553D5F063AD4808ACC7AE9F3375BD35745B0C94C9DD532`

| Field | Value |
|---|---|
| Class | patterned |
| Owner | Rust |
| Files | 2 |
| Depends on | none |
| Parallel with | none |

## Scope

Add the three new `AppSettings` fields (`restart_resume_orchestrator_prompt`, `restart_resume_agent_prompt`, `restart_resume_wake_working_agents`) with their defaults and serde attributes, update the frozen `EXPECTED_NON_PROJECT_SETTINGS_JSON` control, and add the pure `persisted_is_working` predicate that mirrors `is_working`.

Phase 1a ends at a green tree where the settings contract and the persisted-working predicate exist and no behaviour has changed. It is the only phase 1b, 2 and 3 depend on.

## Rules

The phase file is the specification and it is frozen by the digest above. It is decision-complete: implement it as written, do not re-derive it, and do not deviate. If any prescribed text turns out to be wrong against the tree, stop and report rather than working around it: a byte change to the phase file invalidates its digest and requires re-certification.

Verification, including the mandatory positive controls (5 mutations) and the dependency-cycle gate, is specified in the phase file's own acceptance criteria. Record the whole suite green on the unmutated tree before applying any mutation.

Branch: `feature/1793-auto-resume-agents-on-restart`, base `9c0ca239567683ec8b151161c3215980692513a7`.

https://claude.ai/code/session_012wXyfih3hfkZLBbnaMnG8b

Issue #1793, opened 2026-09-05
Auto-resume agents that were working when the app restarts, with configurable injected prompts


mblua:
## Situation

When AgentsCommander is closed, or force-closed, and started again, the sessions come back but the agents do not resume their work. The user has to open every agent and type something by hand so it continues.

What happens today:

- `should_wake_on_restore` / `restore_session_should_wake` (`src-tauri/src/lib.rs:966`, `:983`, called from the restore loop at `:2052`) only decides whether a persisted session comes back with a live PTY or as a dormant `Exited(0)` record. It wakes orchestrators only, gated on `AppSettings::restore_coordinator_wake_state` (`src-tauri/src/config/settings.rs:336`). Non-coordinator replicas are always deferred under the current policy.
- A woken session re-spawns its coding agent with `--continue` (`src-tauri/src/config/sessions_persistence.rs:1911`), so the conversation is restored, but the agent then sits at an empty prompt. Nothing is ever written to it, so the work that was in flight never resumes.

The state needed to do this automatically is already persisted:

- `sessions.json` stores `status` and `waiting_for_input` per session (`PersistedSession`, `src-tauri/src/config/sessions_persistence.rs:437`, `:443`) and is rewritten on every busy/idle edge by the steady-state drivers in `lib.rs`, so even after a hard kill the last real state is on disk. `status` is already consumed on restore.
- `activity.jsonl` (`src-tauri/src/config/activity_log.rs`) is an append-only log that writes a `busy`/`idle` record at the instant of the edge, plus an `app_alive` heartbeat every 60 s carrying `workingSessionIds`, the exact list of sessions that were working (`src-tauri/src/pty/idle_detector.rs:580`, `HEARTBEAT_TICKS = 120` at `CHECK_INTERVAL = 500 ms`).

So the missing piece is not the record. It is the actuation after restore.

## Why this change is needed

Restarting the app is routine: updates, hangs, and normal restarts all happen. Every restart currently costs the user a manual pass over the whole fleet to re-activate agents that were mid-task, and an agent that is never re-activated silently stops making progress. The app already knows which agents were working; it should act on that.

## How it will be changed

**1. Two configurable prompts, injected after restore.**

Add two settings, both editable by the user:

- Orchestrators that were working at shutdown. Default: `AgentsCommander was restarted. Continue with the work that was in flight.`
- Non-coordinator agent replicas that were working at shutdown. Default: `.`

[...]

PR #1872, opened 2026-09-08
feat(#1802): phase 1b - wake policy, target collection and the restart auto-resume pass

### LINKED ISSUE #1802: #1793 phase 1b: replica wake policy, readiness and prompt injection
Child phase of epic #1793.

**Phase file (frozen):** `plans/1793-auto-resume-agents-on-restart/phase-1b-wake-policy-and-injection.md`
**Phase-SHA256:** `1F2034085288B4C842790AAE9D27255D698BE2BB28BEABCFC542416FFEB4E18A`
**Epic Plan-SHA256:** `F11C050377EE0CA444553D5F063AD4808ACC7AE9F3375BD35745B0C94C9DD532`

| Field | Value |
|---|---|
| Class | design-bearing |
| Owner | Rust |
| Files | 3 |
| Depends on | 1a |
| Parallel with | 2, 3 |

## Scope

The behaviour half of the feature: the replica wake arm (`should_wake_working_agent_on_restore`, ORed with the unchanged `should_wake_on_restore` at its single call site), target collection in the restore loop, the readiness rule, the pure `partition_restart_resume_ready` selection function, and the post-restore pass that types the configured prompt into each target once.

This is the only design-bearing phase. Two points carry the whole safety argument and are settled in the phase file:

- **Readiness needs both conjuncts.** `waiting_for_input == true` with `status` not `Exited(_)` is not sufficient on its own: it cannot distinguish an agent at its prompt from an agent that has printed nothing since spawn. The second conjunct is `IdleDetector::has_printable_activity_since` against a `collected_at` stamped in the restore loop. Do not drop it.
- **The helper lives in `commands/pty.rs`, not `commands/session.rs`.** `commands::session -> pty::idle_detector` does not exist; adding it would break the byte-identical `module-arcs.txt` gate. That placement is a decision, not a preference.

## Rules

The phase file is the specification and it is frozen by the digest above. It is decision-complete: implement it as written, do not re-derive it, and do not deviate. If any prescribed text turns out to be wrong against the tree, stop and report rather than working around it: a byte change to the phase file invalidates its digest and requires re-certification.

Verification includes 8 mandatory mutations, each applied alone to a clean tree, each required to die at a bare assertion inside the named test with the panic line recorded; the dependency-cycle gate in its executable detector-then-recorder form; and a manual force-close-and-restart check on a built app. Record the whole suite green on the unmutated tree first.

The task score raised a verification-difficulty veto (14), so the proof itself is reviewed, not only the code.

Branch: `feature/1793-auto-resume-agents-on-restart`, base `9c0ca239567683ec8b151161c3215980692513a7`.

https://claude.ai/code/session_012wXyfih3hfkZLBbnaMnG8b

mblua:
Phase 1b of epic #1793, the behaviour half of the auto-resume feature.

Phase 1a (#1801) landed the settings contract and the `persisted_is_working` predicate with nothing consuming them. This phase adds the replica wake arm, target collection in the restore loop, the readiness rule, the pure `partition_restart_resume_ready` selection function, and the post-restore pass that types the configured prompt into each target once.

## Scope

Three files, 728 insertions, 13 deletions:

- `src-tauri/src/commands/session.rs`
- `src-tauri/src/commands/pty.rs`
- `src-tauri/src/lib.rs`

`src-tauri/module-arcs.txt` is byte-identical to its blob at the base: this phase adds no module and no module-to-module arc.

## The two decisions that carry the safety argument

**Readiness needs both conjuncts.** `waiting_for_input == true` is not sufficient on its own: it means only "no PTY bytes for 2.5 s" and cannot distinguish an agent sitting at its prompt from one that has printed nothing since spawn. The second conjunct is `IdleDetector::has_printable_activity_since` against a `collected_at` stamped in the restore loop, so an agent that has emitted nothing can never read ready. The discriminator can produce a false NOT-ready, never a false ready.

**The readiness helper lives in `commands/pty.rs`, not `commands/session.rs`.** `commands::session -> pty::idle_detector` does not exist among that module's 44 out-arcs, and adding it would break the byte-identical arc-record gate. `commands::pty` already owns that arc and is reached over the pre-existing `commands::session -> commands::pty` arc.

## Plan

Implemented from a frozen phase file that reached three-party consensus over ten rounds.

`plans/1793-auto-resume-agents-on-restart/phase-1b-wake-policy-and-injection.md`
Phase-SHA256 `3DC1152F31AD833EDC876C1717D4A1B4BD82EC90A85200913F6972359004DB64`, 48113 bytes.

The phase's own precondition ran on its first real use and did real work: `main` had moved since certification, so the tree hash no longer matched, and the fallback command over the twelve declared consumed-contract paths printed nothing, which is the condition for proceeding without re-anchoring.

## Verification

Eight mutations, each applied alone to a clean tree, each required to die at a bare assertion inside its named test with the panic line read back against the committed file. The adversarial reviewer re-derived every panic anchor independently and confirmed all eight, checked that M8's fixture is not vacuous, that M3's oracle discriminates because the two class prompt defaults differ, and that M5's table is exhaustive because `SessionStatus` has exactly four variants.

Also verified by the reviewer rather than read: the dependency-cycle gate re-run to a scratch path, regenerating a record byte-identical to the committed one; `cargo fmt --all -- --check` green; `should_wake_on_restore`, its nine assertions and the three preserved `lib.rs` functions identical across both blobs; and all thirteen deleted lines accounted for.

**Stated plainly rather than implied:** the reviewer did not re-run `cargo clippy --workspace --all-targets -- -D warnings` or the full `cargo test`, because that is a cold workspace build on a box that had already out-of-memory-killed a run, and it was instructed not to disturb the application hosting its session. The implementer ran both green. This PR's CI is what settles it on clean runners, which is the stronger evidence.

## Not yet done

Acceptance criterion 7 is a manual check: force-close the built app and restart it, then read the `[restart-resume]` lines from `app.log`. It has not run. It cannot be run from inside this environment, because the application it would force-close is the one hosting the agents doing the work. **This PR is not to be merged until that check passes.**

Closes #1802.


mblua:
## Acceptance criterion 7: run, with the evidence and the residual gap

AC 7 is the manual behaviour check. It has now run, on a build of `332b8a228e8283b1d4d2f2638f923cc77685be5d` produced by `ac-shipper-v4` as `agentscommander_testeable.exe` (SHA-256 `D948E2F3B696EDA2C4549E71E5C7CC87413DF0FB6B3FE2EA73AEAE893ADEEC51`), launched against the live configuration directory via `AGENTSCOMMANDER_CONFIG_DIR`.

It was not a staged fixture. The build ran as the real application across the author's live rooms, three times.

### Observed

| App start | Sessions evaluated | Targets | Final line |
|---|---|---|---|
| 2026-09-07 23:23:11 | 27 | 3 | `[restart-resume] complete: 3 injected, 0 timed out` |
| 2026-09-08 00:02:11 | 30 | 2 | `[restart-resume] complete: 2 injected, 0 timed out` |
| 2026-09-08 00:06:29 | 30 | 1 | `[restart-resume] complete: 1 injected, 0 timed out` |

The first start injected into `room-18-ac-dev-team-v4/ac-tech-lead-v4`, an orchestrator, and into `room-12-ac-dev-team-v4/ac-architect-v4` and `room-18-ac-dev-team-v4/ac-dev-rust-grinch-v4`, both non-coordinator replicas. **Both classes fired**, so `restartResumeWakeWorkingAgents` is exercised on its enabled arm.

What the three runs establish:

- **The settle is honoured.** From the `N target(s); settling 30s` line to the first `injected into` line: 32.6 s, 32.5 s and 37.6 s. All above the 30 s `RESTART_RESUME_SETTLE`.
- **Sessions that were not working receive nothing.** 27 evaluated against 3 targets, then 30 against 2, then 30 against 1. 24, 28 and 29 sessions respectively were evaluated and not collected.
- **The latch holds.** Three app starts, exactly one `[restart-resume]` block each, each opening with its own `N target(s)` line. A fourth start at 2026-09-08 09:36:18 evaluated 30 sessions and produced **zero** `[restart-resume]` lines, because that start was a binary without this change: a clean negative control that arrived by accident.
- **No failure path was taken.** Across all three runs there is no `never reached its prompt` line and no `prompt rejected by the PTY text validator` line. Every target that was collected was injected into.

### The residual gap, recorded rather than glossed

AC 7's second scenario names two sub-cases this evidence does not isolate:

1. A session with `start_fresh_on_restore` set **that was also working** being excluded from collection.
2. The `restartResumeWakeWorkingAgents = false` variant, where the replica stays dormant and receives nothing while the orchestrator still receives its line.

Neither was staged. The bulk-exclusion figures above are consistent with both but do not isolate either, because the excluded population is dominated by sessions that were simply not working.

What stands behind them instead: `restart_resume_prompt_for` returns `None` on `start_fresh_on_restore || !persisted_working`, and the wake arm is gated on the flag; both are covered by the phase's mutation battery, whose eight mutations each died at a bare assertion inside a named test, independently re-derived by the adversarial reviewer.

This is stated as a known limit of the manual check, not as a claim that it was verified. Anyone reopening this behaviour should start by staging those two sub-cases.

### Note on where this ran

The injections went into live agent sessions in rooms 12, 14 and 18. That is stronger evidence than a fixture, and it is also a real side effect: three agents received a resume prompt from a test build. Recording it so the provenance of those nudges is not a mystery later.

PR #1878, opened 2026-09-08
feat(settings): add the On app restart section to Settings > General (refs #1803)

### LINKED ISSUE #1803: #1793 phase 2: On app restart section in Settings > General
Child phase of epic #1793.

**Phase file (frozen):** `plans/1793-auto-resume-agents-on-restart/phase-2-frontend-settings.md`
**Phase-SHA256:** `8D604926B12B934DDDFD22922E9BCF829071C2F86808132AA070B2CA94548847`
**Epic Plan-SHA256:** `F11C050377EE0CA444553D5F063AD4808ACC7AE9F3375BD35745B0C94C9DD532`

| Field | Value |
|---|---|
| Class | patterned |
| Owner | frontend |
| Files | 9 |
| Depends on | 1a |
| Parallel with | 1b, 3 |

## Scope

The Settings UI. A new `On app restart` section in Settings > General, inserted between `Window` and `Orchestrator idle`, containing the existing `restoreCoordinatorWakeState` checkbox moved verbatim out of `Window`, the new replica wake checkbox, and the two prompt inputs. Plus the three new fields on the `AppSettings` TypeScript interface and the seven test fixture builders that must be updated.

No CSS is added: every class the phase uses already exists.

## Rules

The phase file is the specification and it is frozen by the digest above. It is decision-complete: implement it as written, do not re-derive it, and do not deviate. If any prescribed text turns out to be wrong against the tree, stop and report rather than working around it: a byte change to the phase file invalidates its digest and requires re-certification.

Two traps the phase file already handles, flagged so they are not mistaken for regressions:

- Adding three required interface fields breaks `npm run typecheck` in seven fixture builders until they are updated. The phase turns that into a positive control: capture the failing typecheck BEFORE fixing the fixtures, which proves the fields are actually required.
- `npm test` can print every test passed and still exit 1, roughly one run in three, from a `SettingsModal` timer outliving its test. Record the clean-tree baseline totals AND exit code before mutating anything; a "kill" that is really that flake is not a kill.

`.tsx` and `.md` files in this repo are CRLF while the plan file is LF, so match on single lines rather than multi-line anchors.

Verification includes 3 mandatory mutations. M1 moves the checkbox back into `Window`: if the suite stays green, the section test was checking presence instead of membership and the phase is not done.

Branch: `feature/1793-auto-resume-agents-on-restart`, base `9c0ca239567683ec8b151161c3215980692513a7`.

https://claude.ai/code/session_012wXyfih3hfkZLBbnaMnG8b

mblua:
Phase 2 of epic #1793. Adds the **On app restart** section to Settings > General: the two checkboxes and the two configurable restart-prompt text boxes, and moves the existing `restoreCoordinatorWakeState` control out of **Window** into the new section.

Closes #1803.

Phases 1a (#1801) and 1b (#1802) are already merged; this is the UI that was missing, so the settings no longer have to be edited by hand in `settings.json`.

## Scope

9 files, 210 insertions, 11 deletions. No Rust, no CSS, no docs. All eight CSS classes the section uses already existed.

## Three-party plan consensus

The plan was certified through 23 rounds and both verifiers returned `PLAN_APPROVED` on the frozen bytes:

```
phase-2-frontend-settings.md = 58F111F0CEEB6B1351C657FFCE7EA4D8F626E474FCE3501326EC60F8EA8F73BE
```

The shipped JSX is not a retyping of the plan: the reviewer extracted the plan's fence and compared it against the committed blob with `cmp`, and the two are **byte identical, newlines included**.

## Evidence

- **Typecheck positive control, captured before the fix**: with only `src/shared/types.ts` edited, `npm run typecheck` exited 2 with exactly **seven TS2322 errors**, one per fixture, each naming `restartResumeWakeWorkingAgents`, and zero naming the two prompt fields. After the seven fixtures, exit 0. Reproduced independently by the reviewer.
- **Suite**: `npm test` green at **181 files / 2201 tests**, exit 0, first attempt, run independently by both the implementer and the reviewer.
- **Mutation battery**: the plan's three mutations applied one at a time to a clean tree, each **killed** by a bare assertion failure at the named line, with a revert gate between. Re-run independently by the reviewer.
- **Other gates**: `test:debt`, `test:classify:self`, `test:report:self`, `check:frontend-dependencies`, `build`, `validate-branch-name` all exit 0. The repo has no frontend lint or prettier script, so nothing reflowed the fence.

## Known coverage gaps, not blocking

The adversarial review found four mutations the battery does not kill, and none is a regression introduced here. The one worth a follow-up: **no test asserts that the moved `restoreCoordinatorWakeState` checkbox still reflects and writes its field**, so `checked={false}` would leave the suite green. The control's correctness is established here by the shipped bytes being identical to the plan's fence and by inspection of the `checked=` / `updateField` / `data-ac-testid` triple, not by a test. A follow-up issue covers it.

## Review

Implemented by `ac-dev-rust-v4`, adversarially reviewed and `APPROVED` by `ac-dev-rust-grinch-v4`, which re-ran every gate and every mutation itself rather than accepting the report.

https://claude.ai/code/session_012wXyfih3hfkZLBbnaMnG8b


mblua:
## Merging with a failing SonarCloud check, and why

Every required check passes. `main`'s ruleset requires exactly one status check, `validate-branch-name`, and it passed in 7s. Every regression job passed on all four platforms, plus `frontend-regression`, `lockfile-drift`, `rust-fmt`, `test-debt`, `windows-release-cli-smoke`, `terminal-snapshot-portable` and both Socket checks.

**SonarCloud is not a required check and it fails on exactly one condition:**

```
new_reliability_rating        OK     1
new_security_rating           OK     1
new_maintainability_rating    OK     1
new_security_hotspots_reviewed OK    100.0
new_duplicated_lines_density  ERROR  33.8  (threshold 3)
```

No new reliability, security or maintainability finding. The single failure is duplicated new lines, and **the duplication is specified by the plan rather than incidental**: seven test fixtures each receive the byte-identical four-line block

```ts
restartResumeWakeWorkingAgents: false,
restartResumeOrchestratorPrompt:
  "AgentsCommander was restarted. Continue with the work that was in flight.",
restartResumeAgentPrompt: ".",
```

which is 28 of the 210 added lines, and the two prompt controls and two checkbox controls in the new section are structurally parallel because the plan fixes their order and their markup. The uniformity is deliberate: the phase file applies one fixture fence identically to all seven builders **so that no test depends on a fixture-specific value**.

De-duplicating would mean editing bytes that the three-party consensus certified and that both verifiers checked with `cmp` against the committed blob. That trade is not worth taking to move one non-required metric, so this merges as-is and the finding is recorded here instead of being silently bypassed.

Coverage gaps found by the adversarial review are tracked separately in #1879.

https://claude.ai/code/session_012wXyfih3hfkZLBbnaMnG8b

PR #1922, opened 2026-09-10
feat(catalog): add built-in coding-agent support switch (#1912)

### LINKED ISSUE #1912: Code-level enable/disable switch for built-in coding agents in the catalog
## Situation

The Welcome dialog ("Let's set up your first Coding Agent") and the Settings preset list show every entry of the coding-agent catalog. The catalog is `src-tauri/resources/coding-agents/agents.default.json` (8 built-ins: claude, codex, hermes, cursor, pi, opencode, antigravity, muse), embedded via `include_str!` in `src-tauri/src/config/coding_agents_catalog.rs`, seeded ONCE into `<project>/.ac/coding-agents/agents.json` and user-owned thereafter. Every reader (IPC `get_coding_agent_catalog` and its WebSocket route, the legacy config-dir read, the embedded self-heal, the `updateCommands` backfill, `agent_update.rs` update plan, and the CLI `coding-agent catalog` / `add --from-catalog` verbs) goes through `load_catalog_for_settings` -> `load_catalog` -> `validate_and_filter`.

There is no code-level switch to enable or disable a built-in agent: an agent is "supported" simply by being present in the JSON. There is no `enabled` field, no platform gate, no Cargo feature, no `VITE_*` flag. Muse's "macOS/Linux host only" description is text only.

## Why a change is needed

The maintainer needs to withdraw support for some built-in agents with an explicit, reviewable control in code. Editing `agents.default.json` cannot do that: because of the seed-once contract, existing installs keep serving the old entries from their already-seeded `agents.json`, and the frontend mirror `FALLBACK_CODING_AGENTS` (`src/shared/agent-presets.ts`) drifts silently (today it already has 7 entries vs 8 in the backend).

## Planned change

Add a single explicit, code-level enable/disable control for built-in coding agents, enforced on the shared read path so that a de-supported agent disappears from every consumer (Welcome, Settings, WebSocket, CLI, update plan) on every install, including those with an already-seeded catalog, without rewriting user-owned files. The control must be pinned by a test to the set of keys in the embedded catalog so a stale or misspelled key cannot silently do nothing. The frontend fallback mirror follows the enabled set. This issue lands the mechanism with every built-in still enabled; which agents get switched off is a separate decision and will be a one-line follow-up.

Scope of this issue: "hide from all consumers" (Reading A). Not in scope: deleting entries from `agents.default.json`, auto-removing already-registered `settings.agents` rows, gating the config-folder masters, platform gating.

Design proposal and feasibility read: room-12 messaging files `20260909-195243-...-coding-agent-support-flag-proposal.md` and `20260909-195812-...-feasibility-agent-support-flag.md`.

mblua:
The built-in coding-agent catalog had no code-level support switch, so previously seeded manifests kept exposing every built-in. Add BUILTIN_AGENT_SUPPORT to filter catalog reads and newly seeded manifests/config masters while preserving existing user files and legacy manifest bytes. All eight agents remain enabled in this change; the frontend fallback also gains the missing Muse entry.

Validation: dev implementation review and independent Grinch review PASS, including the merge of current main. Module 53, WebSocket 1, CLI 2, frontend 19; merge-overlap suites 66/35/213/46 and 16 Muse tests passed. Formatting, clippy with warnings denied, and TypeScript typecheck passed. Positive controls exercise disabled rows, seed preservation and the steady-state project-lock path. A pre-existing Windows file-access test flake was observed in earlier repeated runs; it did not recur in final merge verification. The unreadable legacy-file seed arm is checked by shared-helper call-site review.

The approved plan is plans/1912-coding-agent-support-switch.md. Final standalone build will follow from merged main.

Closes #1912


mblua:
CI failure investigation on 21f8f93d: the initial Windows runs failed different tests (context_alerts on the PR run, selection on push); each passed in the other run on the same SHA. The affected files match main. Dev reproduced 50/50 context-alert cases, 20/20 selection cases and three complete lib-suite passes (4189 each), and found related selection-test failures on an independent documentation-only branch. Scheduling races remain hypotheses; no #1912 correction is supported by the evidence.

One controlled rerun of each failed Windows job is in progress. Merge remains pending required-check success. Sonar duplication is non-required and corresponds to repeated fields in the Muse fallback data row mandated by the approved plan; retained without changing the catalog design.

PR #2216, opened 2026-09-19
docs(config-seed): note the built-in support gate on factory masters and re-seed (refs #2146)

### LINKED ISSUE #2146: docs: config-seed page omits the BUILTIN_AGENT_SUPPORT gate on factory masters and the Re-seed button
`docs/features/config-seed.md` describes the tier-5 factory default as existing for `.claude`, `.codex` and `.opencode`, and describes the Re-seed button as gated on the command's executable basename. Both are accurate as far as they go, but neither mentions the additional `BUILTIN_AGENT_SUPPORT` gate (#1912) that sits in front of both.

## The missing condition

`supported_embedded_masters()` filters `EMBEDDED_SEED_MASTERS` through `is_supported_builtin` against the table in force:

- `src-tauri/src/config/coding_agents_catalog.rs:3767` — the filter.
- `src-tauri/src/config/coding_agents_catalog.rs:86` — `BUILTIN_AGENT_SUPPORT`.
- `src-tauri/src/config/coding_agents_catalog.rs:287` — `is_supported_builtin`: `false` only for a key present in the table with a `false` row.

That iterator is what feeds every master-related path: `ensure_seeded` writing the masters (`:3867`), `embedded_master_for_command_basename` (`:3793`), and `reseedable_command_basenames` (`:3803`), which is what enables the Re-seed button.

So a built-in whose key carries a `false` row ships no master and gets no Re-seed button, regardless of its Config folder value. The module header at `:31-37` already states this ("the SEED GATE on ... the config-folder masters"); the user-facing doc does not.

## Severity: latent, not currently observable

Worth being explicit so nobody over-scopes the fix. In `BUILTIN_AGENT_SUPPORT` today, `claude`, `codex` and `opencode` are all `true`, and the only `false` row is `muse`, which ships no master. **The gate therefore changes nothing a user can observe right now.** This is documentation of a real code path that becomes user-visible only if one of those three keys is ever switched off, not a description of present-day behavior that is wrong.

## Proposed change

One sentence in `docs/features/config-seed.md`, in the "The factory default and the Re-seed button" section, noting that a built-in disabled in the support table ships no factory master and shows no Re-seed button. Phrase it as a condition on the mechanism, not as a caveat about the current three agents, since all three are enabled.

## Out of scope / explicitly rejected

While reviewing I also checked whether the tier table's `<workspace>/coding-agents/_seed/<dest>` entry was inconsistent with the actual path. **It is not.** The doc defines `<workspace>` as the project's `.ac` root, which matches `master_dir_for_dest` (`:3788`) joining `catalog_dir(ac_dir)`. No change needed there.

## Acceptance

- The support-table gate is stated once in `docs/features/config-seed.md`.
- No code change.
- A domain agent verifies the claim from a cold start against `coding_agents_catalog.rs` before the change lands.

mblua:
Closes #2146.

Documentation only: one sentence in `docs/features/config-seed.md`.

The page describes the tier-5 factory master and the "Re-seed default configuration" button as existing for `.claude`, `.codex` and `.opencode`, gated on the command's executable basename. Both statements are accurate, but neither mentions the `BUILTIN_AGENT_SUPPORT` gate (#1912) that sits in front of both.

`supported_embedded_masters()` (`src-tauri/src/config/coding_agents_catalog.rs:3767`) filters `EMBEDDED_SEED_MASTERS` through `is_supported_builtin` against that table (`:86`). It feeds every master-related path: writing the masters (`:3867`), `embedded_master_for_command_basename` (`:3793`) and `reseedable_command_basenames` (`:3803`), which is what enables the button.

The added sentence states the condition and immediately bounds it: `claude`, `codex` and `opencode` are all enabled today and the only `false` row is `muse`, which ships no master, so the gate changes nothing observable right now. It is documented because it becomes user-visible if any of those keys is ever switched off.

The issue also recorded a second candidate finding — a suspected inconsistency in the tier table's path — and explicitly rejected it: the page defines `<workspace>` as the project's `.ac` root, which matches `master_dir_for_dest`. No change was made there.

## Process note

Authored and landed by the tech-lead at the user's explicit direction, bypassing `docs-implementation-workflow`'s separate author and cold-start verifier, and admin-merged without waiting for PR CI. The code facts above were re-verified against `main` before writing. Recorded here so the deviation is visible in the audit trail.


mblua:
Cerrado sin mergear por decisión del usuario: la redacción la hace `technical-writer` siguiendo `docs-implementation-workflow`, con verificador en frío separado. El texto de este PR queda descartado a propósito para no entregarle prosa terminada que solo transcriba. #2146 sigue abierto.

Issue #2347, opened 2026-09-21
Restore the main window size and maximized state on startup


mblua:
## Situation

A maximized main window comes back small. User report: the app starts at the centered default
(~1400x900, "60x60%" of the screen), they maximize it, use it, close it, and the next start is the
centered default again, so they re-maximize every time.

Verified on `037cf5e2a626cd72d180757d034ba2d2d56243da`:

- Startup reads `mainGeometry` (x/y/width/height, physical px), validates it against the live
  monitors, converts it to logical pixels and builds the `main` window at those bounds; when nothing
  is saved or the rect is judged off-screen it falls back to the centered
  `min(screen_w,1400) x min(screen_h,900)` default (`src-tauri/src/lib.rs:4097-4410`).
- The renderer saves `outerPosition`/`outerSize` 500 ms after each move/resize, as a whole
  `AppSettings` write (`src/shared/window-geometry.ts:13-59`, wired at `src/main/App.tsx:1408`).
- No maximized/fullscreen/minimized state is persisted anywhere: `WindowGeometry` is
  `{x, y, width, height}` (`src-tauri/src/config/settings.rs:226-231`) and no persisted `maximized`
  field exists.
- A window maximized at close saves the *maximized* rect. The pre-maximize restore rect exists only
  in `WINDOWPLACEMENT.rcNormalPosition`, which production code never reads.
- Teardown cancels the pending 500 ms timer instead of flushing it, and the quit path never forces a
  geometry save (`window-geometry.ts:55-59`, `src/main/App.tsx:1457-1475`,
  `src-tauri/src/lib.rs:2629-2635`).
- `mainGeometry` is protected by nothing in `build_protected_settings_candidate`, so a stale
  whole-settings payload can resurrect an older rect
  (`src-tauri/src/commands/config.rs:672-730`). The watchers window already replaced that shape with
  a dedicated one-field command for the same race (#1171).
- If `settings.local.json` carries `mainGeometry`, that value wins on every load and the app can
  never update the effective value (overlay `restore_base` / `reapply_from`).
- `docs/reference/settings.md:290` documents close-time geometry writes that the code does not
  perform.

## Why the change is needed

The user must re-maximize on every start. Persisted geometry alone cannot express "was maximized",
and a special-state rect can replace the useful restore rect; the save can also be lost or regressed
by a later writer.

## How it will be changed (behavioural)

- Persist, for the main window, the last usable normal (non-maximized, non-fullscreen,
[...]

PR #2359, opened 2026-09-21
fix(2348): persist and apply the main window placement pair

### LINKED ISSUE #2348: Persist and apply the main window's normal bounds and display state
Parent epic: #2347
Phase: P1-rust-window-state (class: design-bearing; owner: ac-dev-rust-v4)
Depends on: none. Parallel with: P2-renderer-observation.

## Situation and change

The main window cannot come back maximized: only `mainGeometry` (x/y/width/height) is persisted, and
the save path is a whole-settings write that any stale payload can regress. Persist the main window's
placement as two facts - the last usable normal (non-maximized, non-fullscreen, non-minimized) bounds
and an explicit display state - and apply them on startup after building the window at the normal
bounds.

This child carries the phase scope only. The decision-complete specification (exact symbols, schema,
migration, write ownership, verification commands) is the phase plan at
`D:\0_repos\AgentsCommander_iac\.ac\plans\issue-2347\<this issue>-<slug>.md`, certified by the
architect, the owning dev and Grinch before implementation.

## Acceptance

- Startup builds the main window at the persisted normal bounds with the existing logical conversion,
  and maximizes it when the persisted state says maximized; no production maximize call existed
  before this phase.
- Minimized and fullscreen are transient: the persisted state never stores them, and restore uses the
  last non-minimized, non-fullscreen state.
- A dedicated mutation updates the window fields under the settings lock; a stale whole-settings
  payload cannot regress them.
- An overlay-owned geometry key behaves as the plan decides (documented pin, or rejected write with a
  visible diagnostic), never a silent write-through.
- Existing minimum size and off-screen validation behaviour are preserved, and a removed monitor still
  yields a usable window.
- The Rust verification command set from the phase plan passes.

## Branch

Created from then-current `main` at implementation time as `<type>/<this issue>-<slug>`.

mblua:
Phase P1 of epic #2347. Closes #2348.

## Scope (exactly four files)
- `src-tauri/src/config/settings.rs`
- `src-tauri/src/commands/window.rs`
- `src-tauri/src/commands/config.rs`
- `src-tauri/src/lib.rs`

## What changed
- `MainWindowDisplayState { normal, maximized }` (lowercase serde, default `normal`) plus the defaulted `AppSettings.mainWindowDisplayState`; `mainGeometry` stays physical outer bounds.
- Central whole-settings `Preserve`: for `mainGeometry` and `mainWindowDisplayState` independently, a present on-disk value wins, while an absent/null disk value never erases a caller value synthesized by the load-time `terminalGeometry` -> `mainGeometry` migration. Whole-settings callers (GUI update/draft, CLI-style reconcile) inherit this centrally; no per-caller placement merge logic.
- New narrow `set_main_window_placement` command (registered on the existing Tauri handler). It holds the settings write guard, rejects with the stable `main_window_placement_overlay_pinned` while the overlay owns either placement key (no memory or base-JSON mutation), otherwise writes both keys as one pair through `MainWindowPlacementWriteMode::Explicit` and publishes the re-decoded result.
- Startup: extracted `is_visible_on_monitors`, `physical_to_logical` and the centered-default calculation to module-level functions; display state resolved once with test-placement precedence (test `maximized=false` suppresses a saved maximize); best-effort maximize after build logs on failure and leaves a usable normal window. `focus_main_window`'s defensive recreate branch is explicitly untouched.

## Evidence (base `34f94e20`, head `daedfd72`)
- `cargo fmt --check` -> 0
- `cargo test --locked -p agentscommander issue_2348 -- --test-threads=1` -> 20 passed, 0 failed
- `cargo test --locked --lib config::settings` -> 244 passed (pinned S6/S27 controls updated only by adding `mainWindowDisplayState`)
- `cargo check --locked --all-targets` -> 0
- `cargo clippy --locked --workspace --all-targets -- -D warnings` -> 0
- layering guards (4 binaries) -> 33 passed
- `npm run build` / `test:debt` / `dup:changed` / `cog:scan` / `validate-no-plans-path` -> all 0
- Module-cycle detector pre vs post: identical cyclic-SCC member sets, `moduleCycles=1`, `functionCyclesCrossModule=0`, zero new/removed module pairs, byte-identical arc records (1139 arcs), `module-arcs.txt` unchanged. The sole `generate_handler!` addition is `commands::window::set_main_window_placement` (existing module, no new arc).
- Baseline debt (present with no changes at base): `commands::session::reader_demand_tests::a_restart_into_the_same_claude_path_reads_only_appended_bytes` fails only in the full parallel `--lib` run.

Rebased onto live `main` (`7650b21e` -> `34f94e20`) after the phase base was cut; drift touched only capture/codex-watcher files and `module-arcs.txt`, and all affected evidence above was refreshed on the rebased head.


mblua:
### Step 9.5 — stale-base triage before landing

`git fetch origin` shows `origin/main` at `e6606c9a` while this PR's head is `daedfd7218ace1c133c0161c49b71c3eb4239966`.

Incoming delta (commits reachable from `origin/main` and not from this head):

- `98cbc49a docs(resource-monitor): document the integral monitor screen`

Files in the delta: `docs/features/resource-monitor.md` only.

**Classification: unrelated to this change.** The delta is documentation-only; this PR modifies four files under `src-tauri/src/` (`config/settings.rs`, `commands/config.rs`, `commands/window.rs`, `lib.rs`). There is no overlap with the PR's changed files, their imports or dependents, no shared documentation target, and no change to build configuration, lockfiles, `.github/**` or CI workflows, so the delta cannot affect the CI results already produced for head `daedfd72`.

Per Step 9.5's unrelated-delta path, no branch update and no fresh CI run are required: this PR lands through the admin merge once the CI of the exact head `daedfd72` is green.

PR #2401, opened 2026-09-22
feat(compact): compact sidebar CSS — overlay, compact row, content removal (#2283)

mblua:
Closes #2283

Phase 3 of epic #2236: the compact sidebar CSS — the collapse/expand control as an out-of-flow rail-wide overlay with per-style computed insets, the compact banner row (row = the toggle alone), the banner stacking context with the explicit activation/toggle/lifted ladder, and removal of the scrollable content from layout and tab order. CSS-on-disk pins only; no component change (phase 9 owns the component and the component tests).

## Evidence

- Base: `0deb37cd42d87ad9cd897e508c7ea08088162073` (`.git/PHASE3_BASE`)
- Dependency proof: phase-1 merge `92c9dc6610419b90f8a1b33c9b8b11afa9350a2d` and phase-6 merge `0deb37cd42d87ad9cd897e508c7ea08088162073` are ancestors of `origin/main`; `--ac-rail-width` present in `variables.css`; the compact `.action-bar` rule present in `sidebar.css`.
- Frozen plan: `.ac/plans/2236-sidebar-compact-toggle/2283-compact-css.md`, `Phase-SHA256: 2E5BE96BAF06F5FFC5080BA5A8D5B5981A9A0D8D7D60D08BEBFD931F8C9719BA` (394 lines); epic `Plan-SHA256: 8843F37651A6E6E8E1C4B2CC1D50B31D491D60CE0E610ACA93DA44857CA91FA2`.
- Commits: `02127f91` (implementation) + `e250582ec27bbd4602ac318e2a8094d734f2fbb6` (census correction round: the lifted set is six names, including the `#2271` Co-managed status dot).
- Diff vs base: exactly the 2 declared paths, `394` insertions, `0` deletions (`sidebar.css` +152, `compact-css.test.ts` +242).
- Grinch Step 9 review: `REVIEW_PASS`, visual app change, digest verified; pins 1-7, 3b and 5b green; no `68px` literal; the `.comanaged` dot rule intact.
- Verification commands: `npx vitest run src/sidebar/styles/compact-css.test.ts` 7 passed; `npm run typecheck` exit 0; `npm test` 225 files / 2859 tests passed; `npm run dup:changed` exit 0 (144 baseline clones); `npm run check:served-paths` "2 rows, 2 literals, consistent". Resolved toolchain: node `v24.13.0`, npm `11.6.2`.

Issue #2283, opened 2026-09-20
Phase 3: Compact sidebar CSS (#2236)


mblua:
Part of #2236. This is Phase 3 — Compact CSS: the overlay geometry, the compact row, content removal. The current sidebar does not provide this phase's behavior. The change and proof are specified in the reviewed phase plan.

**Owner:** `ac-dev-webpage-ui-v4`  
**Class:** design-bearing  
**Depends on child issues:** #2277, #2282  
**Files:** 2 (cap 10) — `src/sidebar/styles/sidebar.css`, `src/sidebar/styles/compact-css.test.ts` (new).

## Objective

Ship the CSS that (a) places the collapse/expand control as a transparent, rail-wide overlay flush at
the true pane edge on whichever side the rail is configured, in all 7 sidebar styles, (b) reduces the
compact Root Agent row to that control alone **without changing its height**, (c) gives the banner an
explicit stacking order so its own controls stay reachable, and (d) removes the scrollable content
from layout and from the tab order while compact — with an on-disk pin for each, every offset
**computed** from the style's own declarations rather than typed.

## Acceptance criteria

1. The control is declared as an out-of-flow overlay, exactly the rail width, flush at the true pane
   edge in both states, on **both** rail sides, in all 7 styles, painting no background or border of
   its own. Flushness is **computed** per style and per side from `-(border-left + margin-left)` and
   `+(margin-left - border-right)`: `0/0`, `-3/0`, `-4/0`, `-9/+7`. Never from padding, never
   symmetric, never a value the CSS happens to carry (pins 3, 4).
2. In compact the row is the control alone: every other banner child is `visibility: hidden`, and
   (epic D26) the halo is the veil tail only, with no white line and no glow (pins 5, 6).
3. The compact rule changes **no** padding, so compact row height equals expanded row height in all 7
   styles. The bytes are pinned (pin 5 absence leg); the pixels are D20 gate 4.
4. The banner is one stacking context (`isolation: isolate`, active or not) and declares a strictly
   increasing ladder, activation `0` < toggle `1` < all five lifted names `2`, so no shipped control
   or `title` tooltip is lost under either overlay; the stylesheet declares exactly those five names
   (pin 3b) and phase 9 test 4d re-derives the same list from the component. `:disabled` on the
   activation overlay is `pointer-events: none`, so right-click while busy still reaches the
   container (pin 3b).
5. The compact toggle rule outranks the left-rail rule by **specificity**, so the control's inline
[...]

Issue #2271, opened 2026-09-20
feat: Co-managed phase 8 - status indicator


mblua:
# Co-managed phase 8 of 10

Parent epic: #2232. Owner: Frontend. Class: design-bearing. Depends on: #2270.
Canonical plan: D:/0_repos/AgentsCommander_iac/.ac/plans/issue-2232/phase-8-indicator.md. Implementation requires approval of the exact plan digest before code work.

## Situation and objective

Add an eighth session-circle state, `comanaged`, red in both themes, lit from the idle
edge without ever showing idle, extinguished when the cycle ends. `exited` wins over it;
it wins over `waiting` and `pending`.

This phase paints the state. It does **not** add the surface that turns the feature on;
that is phase 9.

## Planned files

| # | File | Change |
| --- | --- | --- |
| 1 | `src/shared/ipc.ts` | widen `onSessionIdle`'s payload type; add `onSessionComanagedState` |
| 2 | `src/shared/session-activity.ts` | add `comanaged` to `SessionActivity`; accept it as an option; insert it into the precedence function |
| 3 | `src/shared/session-activity.test.ts` | extend the totality and precedence pins |
| 4 | `src/sidebar/components/session-status.ts` | add `comanaged` to `SessionDotClass` and `DOT_CLASS`; thread the option through `sessionDotClass` |
| 5 | `src/sidebar/components/session-status.test.ts` | extend the mapping pin |
| 6 | `src/sidebar/stores/sessions.ts` | the `comanagedBySessionId` sidecar map and its setters |
| 7 | `src/sidebar/App.tsx` | both event handlers |
| 8 | `src/sidebar/components/SessionItem.tsx` | pass the option at the dot (`:335`); `data-ac-comanaged` and `title` on the same element |
| 9 | `src/sidebar/components/ProjectPanel.tsx` | the replica dot (`:2610`, class from `replicaDotClass` at `:2477`): `data-ac-comanaged` and `title`; **and** thread the flag into `sessionEffectiveStatusSearchText` (`:138-140`) and its call site (`:1098-1102`), per 4.2 |
| 10 | `src/sidebar/components/RootAgentBanner.tsx` | the dot (`:390`, class from the `dotClass` memo at `:68-70`): pass the option, `data-ac-comanaged` and `title` |
| 11 | `src/sidebar/components/replica-dot.ts` | pass the option (`:11`) |
[...]

PR #2407, opened 2026-09-22
feat(loops): notice when saving rebaselines the schedule (#2288)

### LINKED ISSUE #2288: Loop modals are silent that a sessionStart-only save rebaselines the schedule
## Situation

Changing a Loop's `sessionStart` through `update_loop` sets `reset_schedule = true` (`src-tauri/src/config/loops.rs:616-618`), so saving that field alone **rebaselines the Loop's schedule**.

Neither `NewLoopModal` nor `EditLoopModal` says so. The copy does not claim the edit is free of scheduling consequences either — it is simply silent.

## Why it is filed rather than fixed in P4

The P4 plan (#2240) is frozen and user-approved, and its strings are part of those approved bytes. The implementer correctly declined to reword anything and raised it instead.

Leaving it silent is also the **consistent** choice today: a `busyCoordinator`-only edit already rebaselines the schedule in exactly the same way and has shipped without a note. So this is not a regression introduced by epic #2233 — that epic only made a second field reach the same path.

## Options

1. Nothing. The behavior matches the existing `busyCoordinator` precedent, and the docs (#2241) state it.
2. A note in both modals covering **every** field that rebaselines the schedule, not just `sessionStart` — otherwise the UI would explain one field and stay silent on its twin.
3. Surface it only when the save would actually rebaseline.

Option 2 or 3 would be a product-copy decision, not a bug fix.

## Coverage in the meantime

The documentation phase (#2241) will state the rebaselining behavior, so it is on the record for users regardless of what the UI does.

Related: #2233, #2240, #2238.

mblua:
## Situation

`update_loop` rebaselines a Loop's schedule (`reset_schedule = true`) when one of six aspects changes: `expr`, `workgroup`, `promptBody`, `busyCoordinator`, `sessionStart`, `enabled` (`src-tauri/src/config/loops.rs:587-625`). Both Loop modals were silent about it, so saving could move the next run — or discard a pending delivery — with no notice to the user.

## Change

- **EDIT** (`EditLoopModal`): shows `Saving this change restarts the schedule. The next run is counted from the moment you save.` only when that save would actually rebaseline. A name-only edit, an untouched form, an edit reverted to the loaded value, or an unloaded baseline show nothing.
- **CREATE** (`NewLoopModal`): always shows `The schedule starts when you create this Loop. The first run is counted from that moment.`
- To keep the condition aligned with the real rule, the patch comparisons move into one pure helper (`buildLoopUpdateInput`) shared with `resetsLoopSchedule`, used by both `handleSave` and the notice memo, so the notice and the payload cannot drift. Scheduler logic is untouched; no `src-tauri/**` change.

## Scope

5 files: `loop-modal-helpers.ts`, `EditLoopModal.tsx`, `NewLoopModal.tsx`, `EditLoopModal.test.ts`, `NewLoopModal.test.ts`. No CSS, no documentation change (the doc correction was explicitly out of scope).

## Verification

- `npx vitest run src/sidebar/components/EditLoopModal.test.ts src/sidebar/components/NewLoopModal.test.ts` → 29 passed
- `npx vitest run` → 227 files, 2884 tests passed
- `npx tsc --noEmit` → exit 0
- `ac-dev-rust-grinch-v4` Step 9 review: **PASS**, with 10 load-bearing mutation probes (each of the six reset keys dropped in turn, an always-true condition, a one-character change in each notice string, and removal of the loading guard): every probe produced failures, all mutations restored, working tree clean.

## Plan and approvals

- Plan: `/home/mblua/0_repos/AgentsCommander_iac/.ac/plans/2288-loop-rebaseline-notice.md`, sha256 `dc228b1331c3c3c3b260957535e5f75770464e7102b36bbd3aac9b75c8d2295a`
- `ac-dev-rust-grinch-v4`: `PLAN_APPROVED`; the user approved the plan and both exact notice strings via Room 26.

The change is user-visible copy, so a visual confirmation of the two modals is tracked separately; it does not alter these bytes.

Closes #2288


mblua:
## Step 9.5 — stale-base triage (branch 3 commits behind `main`)

Notification: `feature/2288-loop-rebaseline-notice` fell behind `main` on 2026-09-22 14:50:57-03:00.

- Branch head: `7f0c840686f42ae0b0c7200136c92bc5dbe6ad64`
- `origin/main`: `df6f918fca88bd509d17b5fd5c56fc28381a1747` (3 commits ahead)
  - `df6f918f` Merge pull request #2401 from `feat/2283-compact-css`
  - `e250582e` fix(compact): lift the #2271 status dot into the banner stacking ladder (#2283)
  - `02127f91` feat(compact): banner collapse/expand overlay and compact row CSS (#2283)
- Delta files: `src/sidebar/styles/sidebar.css` (+152), `src/sidebar/styles/compact-css.test.ts` (+242, new)

**Classification: unrelated.** No branch update and no fresh CI run; the green result of head `7f0c8406` stays valid for landing.

Evidence:

1. **No file overlap.** This PR changes only `src/sidebar/components/EditLoopModal.tsx`, `EditLoopModal.test.ts`, `NewLoopModal.tsx`, `NewLoopModal.test.ts` and `loop-modal-helpers.ts`. The delta touches only `src/sidebar/styles/`.
2. **No selector overlap.** Every added rule is scoped to `.root-agent-banner*`, `.sidebar-layout.sidebar-compact*` or `.sidebar-compact .sidebar-scrollable`. None of those classes is referenced by the changed components (grep: no matches).
3. **No class overlap with the notices.** The new text reuses `.entity-textarea-meta` / `.entity-textarea-hint`; the delta contains no `entity-textarea` selector and no `modal` selector (grep: no matches).
4. **No build inputs affected.** The delta changes no dependency, lockfile, `.github/**`, CI workflow or build/test configuration. The new `compact-css.test.ts` targets only the banner/compact selectors.

Decision (code-implementation-workflow Step 9.5, unrelated-delta path): admin merge as soon as every required check on head `7f0c8406` finishes green; no branch update, no CI re-run.

CI status at triage time: 30 checks pass, 6 pending (macOS only: `rust-regression-macos`, `terminal-snapshot-portable (macos-15)`, `terminal-snapshot-portable (macos-15-intel)` across the two workflow runs), 0 failures.

Issue #2382, opened 2026-09-22
testability: Linux stale-session ui_automation checks time out (pid_is_alive hardcoded true)


mblua:
# testability: Linux stale-session ui_automation checks time out (pid_is_alive hardcoded true)

## Situation

On Linux, two integration checks in `src-tauri/tests/cli_ui_automation.rs` fail deterministically:

- `dead_session_pid_reports_stale_session` — expected `automation_session_stale`; the CLI returned `{"error":"timeout"}`.
- `stale_prior_session_with_running_daemon_reports_automation_not_enabled_on_stdout` — expected `automation_not_enabled`; received `{"error":"timeout"}`.

Isolated run on `main` `facbc69e25265031aa8b7a03c4a542ba029dfa8b` (`cargo test --locked --test cli_ui_automation`): 25 passed, 2 failed; a repeat reproduces exactly the same two, so it is not load flakiness.

Cause: off Windows the liveness probe is a stub — `src-tauri/src/testability/ui_automation.rs:2056` defines `#[cfg(not(target_os = "windows"))] fn pid_is_alive(_pid: u32) -> bool { true }`. A dead PID is never reported dead, so the stale-session branches are unreachable and both checks run until their deadline.

## Why it matters

The workspace sweep cannot be green on Linux: `cargo test --locked --lib --bins --tests --no-fail-fast` reports `cli_ui_automation` 25/2 in every run. This surfaced while closing epic #2287 (its integrated sweep): the epic's own 13 files all pass, and the epic plan explicitly excludes `cli_ui_automation` (it already serializes its copy/spawn), so this platform gap is tracked here instead of expanding that epic. The change will address the Linux behavior of the stale-session detection path so the affected checks can complete; the concrete approach is decided in planning.

## Evidence

- Source `main` SHA: `facbc69e25265031aa8b7a03c4a542ba029dfa8b`
- Commands and full logs: `.ac/room-17-ac-dev-team-v4/room-shared/i2321-epic-final/report.md`, `t2b-tests-nofailfast.log`, `t5-ui-automation-isolated.log`, `t2b-targets.txt`
- Both test files and `src-tauri/src/testability/` are unchanged since the pre-epic baseline `53f3b4edd1f744a56b89fe1e800c1452b56cf83f` (`git log` over that range is empty).

mblua:
## Task technical score

Scored with the `task-technical-scoring` rubric (ten categories, 1-20 each; process ceremony excluded).

[...]

PR #2415, opened 2026-09-22
fix(#2394): real Unix liveness probe for daemon.pid

### LINKED ISSUE #2394: testability: Linux daemon.pid liveness check is stubbed (stale-daemon warning never fires)
## Situation

`src-tauri/src/config/daemon_pid.rs:124-136` gives `is_pid_alive` a non-Windows stub that returns `true` unconditionally. Every daemon-state decision therefore reads a stale `daemon.pid` as `Running` on Linux and macOS: `detect_daemon_state()` is consumed by `cli/list_sessions.rs:114`, `cli/session_safety.rs:17`, `testability/ui_automation.rs:1516`/`:1550` and `config/activity_log.rs:682`. The user-visible stale-daemon warning never fires off Windows.

The stub's comment is also stale: it justifies itself by saying `src-tauri/Cargo.toml` does not list `libc`, but `libc = "0.2"` is now a direct `cfg(unix)` dependency (`Cargo.toml:45-46`), so the stated obstacle no longer exists.

## Why it matters

On Linux the CLI cannot distinguish a live daemon from a stale snapshot, so it never warns about a dead `daemon.pid` and keeps reporting `Running` for a dead PID. This is the same platform-parity gap that #2382 fixes for the UI-automation session check; #2382 deliberately left this file out because it is a different caller with its own user-visible effect.

## How it could change

Give the non-Windows branch a real liveness probe (mirroring #2382's Unix implementation, or sharing it), and correct the stale `libc` comment. The concrete approach belongs to this issue's own planning: `is_pid_alive` here is private with several callers, so the plan must state whether to reuse the #2382 helper or keep the two independent.

## Evidence

- `src-tauri/src/config/daemon_pid.rs:124-136` — non-Windows stub and its stale `libc` comment
- Call sites listed above (CLI session listing, session safety, UI automation, activity log)
- #2382 lands the first real Unix liveness probe for a sibling caller

mblua:
Closes #2394

Non-Windows `is_pid_alive` in `config/daemon_pid.rs` now delegates to the Unix probe from #2382 (`testability::ui_automation::pid_is_alive`) instead of returning `true`; stale `libc` comment fixed. Windows unchanged. Only `src-tauri/src/config/daemon_pid.rs` changes.

- Plan: `.ac/plans/2394-daemon-pid-liveness.md`, Plan-SHA256 `92a6284273facd1451457a675600ef4843fcc6bec6536a54798a72201e51133e` (grinch PLAN_APPROVED, user-rep APROBADO)
- Grinch code review: PASS on 3a4d87f0
- Linux: fmt, clippy -D warnings, daemon_pid 8/8, pid_is_alive 2/2, reaped_pid 1/1
- macOS/Windows: verified by CI only


mblua:
Step 9.5 delta triage: branch is 9 commits behind main. Delta files: `src-tauri/src/config/settings.rs` (+3/-3) and frontend files under `src/shared/**`, `src/sidebar/components/**`. Classification: **unrelated**. No overlap with the PR file `src-tauri/src/config/daemon_pid.rs`, which imports only `std::path` and does not use `settings`; no build config, lockfile, `.github/**` or CI workflow in the delta. Landing with admin merge once the current head's CI is green, without updating the branch.

Issue #2411, opened 2026-09-22
CI notice wakes an auto-closed orchestrator into a fresh session without its context


mblua:
## Situation

When a room orchestrator pushes a branch and waits for CI, the run can take close to an hour. During that wait the orchestrator is idle, and the auto-close feature (`session/auto_close.rs`) terminates its session.

When CI finishes, the remote-activity notifier (`session/remote_alerts.rs` -> `MailboxPoller::deliver_internal_system_notice` in `phone/mailbox.rs`) already wakes the orchestrator:
- live session -> settle and inject;
- `Exited` session -> destroy it and respawn **with resume**;
- no session at all -> background spawn **without resume**.

Auto-close removes the session entirely, so the third branch runs. The notice is delivered to a brand-new coding-agent session that has none of the prior conversation. It does not know what it was doing, which branch it pushed, or what the next step is.

## Evidence (app.log, local time UTC-3, Room 21, branch `feat/2270-routing-and-provenance-queue`)

- `2026-09-21 22:34:22` `[auto-close] terminated idle session 6a52812c` (Room 21 orchestrator).
- `2026-09-22 00:23:48` `[session] resume-decision dc4eb7da agent=pi cwd=".../room-21-ac-dev-team-v4/__agent_ac-tech-lead-v4" skip_auto_resume=true`, spawned right after a `[RemoteSweeper] round`.
- `2026-09-22 00:23:57` `[remote-activity] delivered coordinatorSession=dc4eb7da ... kind=CiFinished ... branch=feat/2270-routing-and-provenance-queue sha=7dd6fb5`.
- Later notices delivered to the still-live session worked normally (`01:46:31` CiFinished `54998f0`, orchestrator reacted within 2 min).
- `2026-09-22 03:33:25` auto-close terminated `dc4eb7da` again.
- No `delivery failed`, `dropped`, or `hourly notice cap` lines for Room 21.

## Why a change is needed

The wake itself works, but an orchestrator woken without its context cannot act on the CI result. The CI-wait handoff is effectively lost whenever auto-close fires first.

## Intended change

Ensure a CI-finished (and CI-started / branch-stale) notice reaches an orchestrator that still has its working context, even when auto-close ran during the wait. The shape of the fix (e.g. auto-close awareness of in-flight CI, resume on the no-session wake path, or another approach) is to be decided by the architect.

## Acceptance criteria

[...]

Issue #2389, opened 2026-09-22
test(capture): de-flake a_baseline_is_consumed_but_never_routed on Windows rust-regression


mblua:
## Situation

On PR #2385 (branch `feat/2271-co-managed-indicator`), the `rust-regression` job failed on a commit that carries **no Rust changes**, and then passed on a later head with the identical Rust tree:

- run `35689980230` (`pull_request`, head `8692ceb5`, job `106624698155`) — **failure**
- run `35696520780` (`pull_request`, head `49b78598`, job `106644629100`) — **success**
- `rust-regression-linux` passed on both heads.

Failure output:

```
thread 'capture::state::tests::a_baseline_is_consumed_but_never_routed' panicked at src-tauri\src\capture\state.rs:1569:14:
a baseline commits, so it can never be offered again: PreconditionsStale
test result: FAILED. 5000 passed; 1 failed; 20 ignored; 0 measured; 0 filtered out; finished in 166.62s
```

So this is a test-side load flake on the Windows leg, not a product regression.

## Why the test can fail

`commit_effect` (`src-tauri/src/capture/state.rs`) checks staleness twice: before taking the lock, and again after the bounded lock wait (`LOCK_WAIT_BUDGET = 100 ms`). Both compare `pre.observed_at.elapsed() > STALE_PRECONDITIONS` (250 ms) and return `AbstainReason::PreconditionsStale`. The test constructs `preconditions(EffectKind::Automatic)` immediately before the call, so any scheduler stall longer than 250 ms between construction and the check turns the `.expect("a baseline commits, so it can never be offered again")` into a panic. A Windows runner executing ~5000 lib tests in parallel can produce that stall.

Precedent for the class: `31b79ecd` "#2267 phase 4: tolerate a starved lock holder in capture test 25".

This test is not in the known-flaky inventory (#1241).

## Why it matters

It turns an unrelated PR's required `rust-regression` check red and costs a full rerun of a 16-50 minute job, and it hides real failures behind rerun noise.

## How it will be changed

The owning Rust dev takes a test-side fix: make this baseline-commit test robust to scheduler starvation while preserving what it pins (a baseline is consumed, never routed, and spends no budget). The staleness semantics themselves must stay covered; `PreconditionsStale` pinning (test 19) must not be weakened, and no production behavior change is expected.

mblua:
Additional instance of the same class, on PR #2387 (`feat/2282-compact-composition`, head `056328c0`, which carries no Rust changes):

[...]

PR #2437, opened 2026-09-23
fix(session): keep close markers when a coordinator create fails (refs #2413)

### LINKED ISSUE #2413: auto_closed_at marker lost when a manual reopen or peer wake fails after the create-path clear
## Situation

`CoordinatorClocks::auto_closed_at(fqn)` marks a room orchestrator that was closed by auto-close. The shared create path (`commands/session.rs`, the coordinator-create clear around :1765-1799 at base 396e565) clears this marker **before** resource admission (:1838) and the PTY spawn (:2687-2693).

If a **manual reopen** or a **peer (inter-agent) wake** fails after that clear, the marker is lost. A later retry treats the orchestrator as never auto-closed.

This behavior already exists today; #2411 does not introduce it. #2411 fixes only the internal-system wake path (CI / context notices): when that spawn fails, it restores the marker with its original timestamp.

## Why it matters

After #2411, a lost marker means a later internal wake starts the orchestrator fresh instead of resuming its conversation. The UI's "closed by auto-close" indicator also disappears even though the session never came back.

## Possible direction (architect analysis, not decided)

Move the clear to after a successful PTY spawn, or restore it on failure for every create path. Moving the clear changes **when the UI indicator updates for every session create**, which is a user-visible change and outside #2411's scope.

## Status

Follow-up of #2411. Not scheduled. Prioritization and any user-visible decision belong to the user.

mblua:
Closes #2413

## What
When a manual reopen, peer wake or internal wake fails after the create-path clear, the `auto_closed_at` / manual close markers are restored with their original timestamps — only when no PTY/runtime was ever launched.

- Failure-aware handoff ledger in `CoordinatorClocks` (in-memory, persistence format unchanged): success wipes escrow; last failing overlapping create restores.
- Spawn fence + backend `LaunchWitness` (local/container): any end after a possible launch never restores → no stale marker on a live session.
- `CLOSE_MARKER_EVENT_ORDER` keeps clear/restore UI events in clock order.
- #2411 mailbox-only restore removed; the shared create path now covers every caller.

Plan (frozen, 3-party consensus round 6, user-approved): `.ac/plans/2413-auto-closed-marker-lost.md`, Plan-SHA256 D9C2C3D0BD068924BA3AF558D868560FBFD0BC922C32B9D862249DA1A02E20A3.

## Review
- ac-dev-rust-grinch-v4: FAIL on 331c3d6 (event ordering race) → fixed in 62ff325 → PASS; merge of main b740e8b focused review PASS.
- fmt/clippy clean; focused modules green. 3 pre-existing Linux failures of Windows-path tests reproduced identically at base b261e3f.

Final build: shipper after merge.


mblua:
**Stale-base triage (Step 9.5)** — head `e01a2aa1` vs `origin/main` `514c6bc4`.

Delta: the new `main` commit is `514c6bc4 test(capture): retry baseline commit only on measured stale preconditions, refs #2389`.
- It changes only `src-tauri/src/capture/state.rs`, and only inside `mod tests` (a retry loop in one test).
- This PR does not touch `capture/state.rs`. No production code, build config, lockfile or workflow changes.

Decision: this change cannot affect this PR, so the branch is not updated. CI on head `e01a2aa1` stays valid evidence.

PR #2438, opened 2026-09-23
test(capture): de-flake a_baseline_is_consumed_but_never_routed (#2389)

### LINKED ISSUE #2389: test(capture): de-flake a_baseline_is_consumed_but_never_routed on Windows rust-regression

mblua:
Closes #2389

Test-only change in src-tauri/src/capture/state.rs (test 13): retry the baseline commit only when the attempt returned PreconditionsStale AND its snapshot measurably aged past STALE_PRECONDITIONS, at most 5 attempts. Stale abstain mutates nothing, so retry is safe. All assertions unchanged; test 19 and production code untouched. This mitigates the Windows load flake; five consecutive >250 ms stalls still fail.

Verification: capture::state 26 passed; fmt and clippy -D warnings clean; mutants M1/M2/M3 fail as required (M3 shows exactly 5 stale attempts), positive control passes with [Err(PreconditionsStale), Ok(false)]. Grinch PASS on 514c6bc4 and on merge 5249a34e.


mblua:
Step 9.5 stale-base triage: main advanced 436255d0 -> 38c6bd03 (c0933f9c, 99b915af, ed6ad530). Delta is frontend sidebar only (src/sidebar/**); main did not change src-tauri/src/capture/state.rs, no build config, lockfile or workflow. Unrelated to this test-only change, reusing the green CI on head 5249a34e.

